PR #25, opened 2023-09-07
Include API project

MikeEdgar:
Adds console-api to the `api` directory, along with license and DCO information from that repository. Also updates actions to release both API and UI components.


riccardo-forina (approved):
LGTM

PR #30, opened 2023-09-08
Return and display multiple offsets, calculate number of records

MikeEdgar:
Sample response with all offsets included for the partition.
```json
{
  "data": {
    "id": "s0QDg0v1Q16m4QqQDvJEPQ",
    "type": "topics",
    "attributes": {
      "name": "__strimzi-topic-operator-kstreams-topic-store-changelog",
      "internal": false,
      "partitions": [
        {
          "partition": 0,
          "leader": {
            "id": 0,
            "host": "broker-0.cluster-a.192.168.39.252.nip.io",
            "port": 443
          },
          "replicas": [
            {
              "id": 0,
              "host": "broker-0.cluster-a.192.168.39.252.nip.io",
              "port": 443
            }
          ],
          "isr": [
            {
              "id": 0,
              "host": "broker-0.cluster-a.192.168.39.252.nip.io",
              "port": 443
            }
          ],
          "offsets": {
            "earliest": {
              "offset": 0,
              "leaderEpoch": 0
            },
            "latest": {
              "offset": 5,
              "leaderEpoch": 0
            },
            "maxTimestamp": {
              "offset": 4,
              "timestamp": "2023-09-07T18:02:17.546Z",
              "leaderEpoch": 0
            },
            "timestamp": {
              "offset": 0,
              "timestamp": "2023-08-23T10:56:53.639Z",
              "leaderEpoch": 0
            }
          }
        }
      ],
      "authorizedOperations": [
        "CREATE",
        "ALTER_CONFIGS",
        "DESCRIBE_CONFIGS",
        "ALTER",
        "DESCRIBE",
        "DELETE",
        "READ",
        "WRITE"
      ]
    }
  }
}
```
Display in the UI
image

### DIFF
--- a/ui/api/bookmarks.ts
+++ b/ui/api/bookmarks.ts
@@ -62,11 +62,16 @@ export async function createBookmark({
 
 export async function getClusters(): Promise<Cluster[]> {
   const url = `${process.env.BACKEND_URL}/api/kafkas?fields%5Bkafkas%5D=name,bootstrapServers,authType`;
-  const res = await fetch(url, {
-    headers: {
-      Accept: "application/json",
-    },
-  });
-  const rawData = await res.json();
-  return Response.parse(rawData).data;
+  try {
+    const res = await fetch(url, {
+      headers: {
+        Accept: "application/json",
+      },
+    });
+    const rawData = await res.json();
+    return Response.parse(rawData).data;
+  } catch (err) {
+    console.error(err);
+    throw err;
+  }
 }
--- a/ui/components/topicPage.tsx
+++ b/ui/components/topicPage.tsx
@@ -269,7 +269,7 @@ export function TopicDashboard({ topic }: { topic: Topic }) {
               case "recordCount":
                 return (
                   <Td key={key} dataLabel={"Record Count"}>
-                    {row.recordCount + '' || '-'}
+                    {row.recordCount ?? '-'}
                   </Td>
                 );
             }
--- a/ui/components/topicPage.tsx
+++ b/ui/components/topicPage.tsx
@@ -261,15 +261,15 @@ export function TopicDashboard({ topic }: { topic: Topic }) {
               case "offsets":
                 return (
                   <Td key={key} dataLabel={"Offset"}>
-                    {row.offsets?.earliest?.offset}
+                    {row.offsets?.earliest?.offset ?? '-'}
                     /
-                    {row.offsets?.latest?.offset}
+                    {row.offsets?.latest?.offset ?? '-'}
                   </Td>
                 );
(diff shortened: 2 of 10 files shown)


ui/components/topicPage.tsx
```diff
@@ -252,10 +258,18 @@ export function TopicDashboard({ topic }: { topic: Topic }) {
...
-                    {row.offset.offset}
+                    {row.offsets?.earliest?.offset}
+                    /
+                    {row.offsets?.latest?.offset}
+                  </Td>
+                );
+              case "recordCount":
+                return (
+                  <Td key={key} dataLabel={"Record Count"}>
+                    { (row.offsets?.latest?.offset ?? 0) - (row.offsets?.earliest?.offset ?? 0) }
```
MikeEdgar, line 272: I'm not sure if this is the ideal syntax to handle optional values, but it seems to work properly.

ui/api/bookmarks.ts
```diff
@@ -62,11 +62,16 @@ export async function createBookmark({
...
+      headers: {
+        Accept: "application/json",
+      },
+    });
+    const rawData = await res.json();
+    return Response.parse(rawData).data;
+  } catch (err) {
+    console.error(err);
+    throw err;
+  }
```
MikeEdgar, line 76: Running locally with the API at `localhost` fails without anything useful logged. Just adding this here to help debug in the future for anyone with the same issue. The solution is to use `127.0.0.1` instead.

MikeEdgar:
Would it be better to have the backend to the `recordCount` calculation?

ui/components/topicPage.tsx
riccardo-forina, line 272:
```suggestion
                    { (row.offsets?.latest?.offset || '-') / (row.offsets?.earliest?.offset || '-') }
```

What about this? It would show `- / -` in case of no data, and eg. `6 / 9` with data

riccardo-forina (approved):

riccardo-forina:
> Would it be better to have the backend to the `recordCount` calculation?

It would be nice, if we plan to show that info somewhere else.

ui/components/topicPage.tsx
MikeEdgar: I added this change to the Offsets table cell. My IDE suggested using `??` instead of `||`.

MikeEdgar:
> > Would it be better to have the backend to the `recordCount` calculation?
> 
> It would be nice, if we plan to show that info somewhere else.

Latest commit adds a calculated `recordCount` for each partition at for the topic. It is also fetchable via the `fields[topic]` parameter and included by default for the describe topic operation.

PR #36, opened 2023-09-11
Cursor-based pagination for list operations

### LINKED ISSUE #29: Flaky API test: RecordsResourceIT.testConsumeRecordsAsOfTimestamp
```
Error:  com.github.eyefloaters.console.api.RecordsResourceIT.testConsumeRecordsAsOfTimestamp(Instant, Instant, Instant, int)[1] -- Time elapsed: 1.593 s <<< FAILURE!
java.lang.AssertionError: 
1 expectation failed.
JSON path data doesn't match.
Expected: a collection with size <2>
  Actual: <[]>
```

### LINKED ISSUE #31: Add pagination to list requests
Originally https://github.com/eyefloaters/console-api/issues/13

MikeEdgar:
Closes #31 
Fixes #29 

Implementation of [cursor-pagination](https://jsonapi.org/profiles/ethanresnick/cursor-pagination) for Kafka cluster and topic list operations.

### DIFF
--- a/api/src/main/java/com/github/eyefloaters/console/api/mapping/InvalidPageCursorExceptionHandler.java
+++ b/api/src/main/java/com/github/eyefloaters/console/api/mapping/InvalidPageCursorExceptionHandler.java
@@ -1,0 +1,38 @@
+package com.github.eyefloaters.console.api.mapping;
+
+import java.util.List;
+
+import jakarta.ws.rs.core.Response;
+import jakarta.ws.rs.ext.ExceptionMapper;
+import jakarta.ws.rs.ext.Provider;
+
+import org.jboss.logging.Logger;
+
+import com.github.eyefloaters.console.api.model.Error;
+import com.github.eyefloaters.console.api.model.ErrorResponse;
+import com.github.eyefloaters.console.api.support.ErrorCategory;
+import com.github.eyefloaters.console.api.support.InvalidPageCursorException;
+
+@Provider
+public class InvalidPageCursorExceptionHandler implements ExceptionMapper<InvalidPageCursorException> {
+
+    Logger logger = Logger.getLogger("com.github.eyefloaters.console.api.errors.client");
+    private static final ErrorCategory CATEGORY = ErrorCategory.get(ErrorCategory.InvalidQueryParameter.class);
+
+    @Override
+    public Response toResponse(InvalidPageCursorException exception) {
+        List<Error> errors = exception.getSources()
+            .stream()
+            .map(source -> {
+                Error error = CATEGORY.createError(exception.getMessage(), exception, source);
+                logger.debugf(exception, "error=%s", error);
+                return error;
+            })
+            .toList();
+
+        return Response.status(CATEGORY.getHttpStatus())
+                .entity(new ErrorResponse(errors))
+                .build();
+    }
+
+}
--- a/api/src/main/java/com/github/eyefloaters/console/api/model/ListFetchParams.java
+++ b/api/src/main/java/com/github/eyefloaters/console/api/model/ListFetchParams.java
@@ -1,7 +1,6 @@
 package com.github.eyefloaters.console.api.model;
 
 import java.io.ByteArrayInputStream;
-import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Base64;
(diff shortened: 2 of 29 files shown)


riccardo-forina (approved):
👍

api/src/main/java/com/github/eyefloaters/console/api/mapping/InvalidPageCursorExceptionHandler.java
```diff
@@ -0,0 +1,38 @@
...
+
+import com.github.eyefloaters.console.api.model.Error;
+import com.github.eyefloaters.console.api.model.ErrorResponse;
+import com.github.eyefloaters.console.api.support.ErrorCategory;
+import com.github.eyefloaters.console.api.support.InvalidPageCursorException;
+
+@Provider
+public class InvalidPageCursorExceptionHandler implements ExceptionMapper<InvalidPageCursorException> {
+
+    Logger logger = Logger.getLogger("com.github.eyefloaters.console.api.errors.client");
```
mstruk, line 19: Looks intentional but at the same time a bit unusual so just to double check - it is intentional that it's a string here and not this class's FQDN.

api/src/main/java/com/github/eyefloaters/console/api/model/ListFetchParams.java
```diff
@@ -24,8 +72,241 @@ with a minus (U+002D HYPHEN-MINUS, "-"), in which case it will be descending.
...
+     * for the sort operation, both implicitly and explicitly (i.e. the ID is always
+     * the final sort field).
+     * </ul>
+     *
+     * This method called once for each `page[after]` and `page[before]` query
+     * parameter present in the request.
+     *
+     * @param cursor one of the request cursors.
+     * @return true if the cursor was successfully parsed and is valid vis-à-vis the
+     *         `sort` query parameter parameter, else false.
```
mstruk, line 220:
```suggestion
     *         `sort` query parameter, else false.
```

api/src/main/java/com/github/eyefloaters/console/api/support/ErrorCategory.java
```diff
@@ -9,12 +10,15 @@
...
+import com.github.eyefloaters.console.api.model.ListFetchParams;
 
 public abstract class ErrorCategory implements Payload {
 
     private static final Map<Class<? extends ErrorCategory>, ErrorCategory> INSTANCES = Stream.of(
             new InvalidQueryParameter(),
+            new MaxPageSizeExceededError(),
             new ResourceNotFound(),
             new ServerError(),
             new BackendTimeout())
```
mstruk, line 24: IDEA gives me an interesting warning here:
`Referencing subclass BackendTimeout from superclass ErrorCategory initializer might lead to class loading deadlock`

mstruk (approved):
A few little things there. Other than that I wonder if using `page[size]`, `page[before]`, `page[after]` is the best choice of parameter names since the square brackets have to be urlencoded, which complicates readability. I personally would use `page.size`, or `page-size`.

api/src/main/java/com/github/eyefloaters/console/api/mapping/InvalidPageCursorExceptionHandler.java
MikeEdgar: The idea here is that all client errors can be configured with a common logger name (there is an equivalent for the 5XX server errors). An alternative would be to move the mappers into sub-packages, which I am not opposed to either.

Let's do it in a separate PR if it seems like a good idea. Minimally, a constant field with the logger names would probably be a good idea though.

MikeEdgar:
> A few little things there. Other than that I wonder if using `page[size]`, `page[before]`, `page[after]` is the best choice of parameter names since the square brackets have to be urlencoded, which complicates readability. I personally would use `page.size`, or `page-size`.

@mstruk , the square brackets come from the [JSON:API spec](https://jsonapi.org/format/) and [Cursor Pagination sub-spec](https://jsonapi.org/profiles/ethanresnick/cursor-pagination) that the API is following. While the encoded URLs don't look pretty, it should be mostly or completely handled by the HTTP libraries. The benefit is that we can skip over thinking about the details of quite a bit about the API's style since it's specified there.

PR #39, opened 2023-09-14
Issue 38

### LINKED ISSUE #29: Flaky API test: RecordsResourceIT.testConsumeRecordsAsOfTimestamp

### LINKED ISSUE #38: Refactor error mappers to client and server subpackages
Refactor error mappers to client and server subpackages and use FQCN-based loggers. 

From discussion in https://github.com/eyefloaters/console/pull/36#discussion_r1325725046

MikeEdgar:
Closes #38 
Fixes #29


api/src/main/java/com/github/eyefloaters/console/api/errors/client/AbstractNotFoundExceptionHandler.java
```diff
@@ -13,7 +13,7 @@
 
 abstract class AbstractNotFoundExceptionHandler<T extends Throwable> implements ExceptionMapper<T> {
 
-    Logger logger = Logger.getLogger("com.github.eyefloaters.console.api.errors.client");
+    Logger logger = Logger.getLogger(getClass());
```
tombentley, line 16: I don't have a strong opinion on this, but will point out: Using `getClass()` will mean that for subclasses the line numbers in the logs won't correlate with the apparent log class that appears in the logs. That can be confusing, both for users and maintainers of the project. 

FWIW Strimzi chose the pattern of `static final Logger LOGGER = Logger.getLogger(AbstractNotFoundExceptionHandler.class)`. That doesn't mean you have to do the same.

api/src/main/java/com/github/eyefloaters/console/api/errors/client/ConstraintViolationHandler.java
```diff
@@ -1,4 +1,4 @@
-package com.github.eyefloaters.console.api.mapping;
+package com.github.eyefloaters.console.api.errors.client;
```
tombentley, line 1: If we're adding subpackages lets add them with a `package-info.java` to make explicit the rationale for what belongs where. That makes it much easer for later contributors to figure out for themselves where any new handler would belong.

api/src/main/java/com/github/eyefloaters/console/api/errors/client/ConstraintViolationHandler.java
```diff
@@ -23,7 +23,7 @@
 @Provider
 public class ConstraintViolationHandler implements ExceptionMapper<ConstraintViolationException> {
 
-    Logger logger = Logger.getLogger("com.github.eyefloaters.console.api.errors.client");
+    Logger logger = Logger.getLogger(ConstraintViolationHandler.class);
```
tombentley, line 26: If you're using a constant expression for the logger name you might want to make the `logger` `static final`?

tombentley (approved):
LGTM, with a few minor comments.

api/src/main/java/com/github/eyefloaters/console/api/errors/client/ConstraintViolationHandler.java
MikeEdgar: Added `package-info.java` files with JavaDoc comments :+1:

api/src/main/java/com/github/eyefloaters/console/api/errors/client/ConstraintViolationHandler.java
MikeEdgar: Changed constant class-based loggers to `static final`.

api/src/main/java/com/github/eyefloaters/console/api/errors/client/AbstractNotFoundExceptionHandler.java
MikeEdgar: I generally would prefer using the parent class for the parent logger, but in this case since the child classes are empty I think it's more helpful to have their names in the log. This should be revisited if anything changes with this class hierarchy in the future.

PR #42, opened 2023-09-18
Retrieve topic/partition storage information using Admin `describeLogDirs`

MikeEdgar:
Topic summary storage:
image

Partition storage:
image


ui/utils/format.ts
```diff
@@ -0,0 +1,17 @@
...
+
+    if (bytes == 0) {
+        return '0 B';
+    }
+
+    let dm = decimals || 2;
+    let i = Math.floor(Math.log(bytes) / Math.log(K));
+
+    return parseFloat((bytes / Math.pow(K, i)).toFixed(dm)) + ' ' + UNITS[i];
+}
```
riccardo-forina, line 17: Nice! Tomorrow I'll try the native format function, it should support bytes from what I see in the doc:

https://next-intl-docs.vercel.app/docs/usage/numbers#formatting-plain-numbers

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#options

https://tc39.es/ecma402/#table-sanctioned-single-unit-identifiers

MikeEdgar: Ok, that would be much better than this that I found on a Github gist :rofl:

ui/utils/format.ts
```diff
@@ -1,17 +1,28 @@
...
+  return function formatBytes(bytes: number) {
+    const res = convert(bytes, "bytes").to("best");
+    const unit: Record<(typeof res)["unit"], string> = {
+      bits: "bit",
+      GiB: "gigabyte",
+      KiB: "kilobyte",
+      GB: "gigabyte",
+      MiB: "megabyte",
+      PiB: "petabyte",
+      TiB: "terabyte",
```
MikeEdgar, line 15: Should these be `"gibibyte"`, `"kibibyte"`, etc.?

riccardo-forina: yes! And no. Or maybe? The intl api doesn't support the iB units, so we have to use the bytes ones. Which is ok, as we have the data in bytes, thus we'll not get the iB variants and those mappings will never be used.

Although one could argue that we should display the data in KiB et friends instead of KB; checking the Wikipedia KB is base 10, where KiB are base 2.

I hate this stuff 🙃

WDYT, should we use KiB or KB?

MikeEdgar: I guess it depends on what users mostly expect. I think the ones that use the traditional prefixes kilo, mega, etc. are sometimes base 10 and sometimes base 2 whereas the iB units are explicitly base 2. This may be something to raise for discussion as a general topic so that it aligns completely within the console and with other components that may be involved in a streaming platform.

MikeEdgar: I see now on the Wikipedia page that the traditional SI units should be limited base 10, but I guess that doesn't prevent it from being ambiguous to individuals still :-) Ultimately, I don't mind whatever way we go as long as it's consistent.

MikeEdgar: Maybe for now we can just remove the iB variants if they'll never be used by this logic as it is. wdyt?

MikeEdgar: Ok, we should definitely come back and change that based on the outcome of the ADR.

riccardo-forina: I changed the code so it's clear what we are doing. We are getting bytes in input, we are outputting them in *iB (so, base 2).

Eg. 10875778 bytes is 10.372 MiB

MikeEdgar: LGTM - so "imperial" means the *iB units? btw, I'm looking at the test that is _still_ flaky somehow.

riccardo-forina: Yeah, TIL. But it's specific to this library, probably it's like so because it uses a common interface for all units that it supports and it started supporting things like distances and weights? Not sure 🤷

api/src/main/java/com/github/eyefloaters/console/api/model/PartitionInfo.java
```diff
@@ -35,39 +36,46 @@ public PartitionInfo(int partition, Node leader, List<Node> replicas, List<Node>
...
         List<Node> replicas = info.replicas().stream().map(Node::fromKafkaModel).toList();
-        List<Node> isr = info.isr().stream().map(Node::fromKafkaModel).toList();
-        return new PartitionInfo(info.partition(), leader, replicas, isr);
+        Integer leader = info.leader().id();
+        List<Integer> isr = info.isr().stream().map(org.apache.kafka.common.Node::id).toList();
+        return new PartitionInfo(info.partition(), replicas, leader, isr);
     }
 
-    static Either<OffsetInfo, Error> offsetOrError(Either<OffsetInfo, Throwable> offset) {
+    static <P> Either<P, Error> primaryOrError(Either<P, Throwable> offset, String message) {
```
tombentley, line 45: The name of the `offset` parameter should be `primary`?

api/src/main/java/com/github/eyefloaters/console/api/model/ReplicaInfo.java
```diff
@@ -0,0 +1,25 @@
+package com.github.eyefloaters.console.api.model;
+
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+public record ReplicaInfo(
+        @Schema(readOnly = true, description = "The total size of the log segments in this replica in bytes.")
+        long size,
```
tombentley, line 7: We need to defined whether this is node-local storage, or total storage, including remote log segments in an object store.

api/src/main/java/com/github/eyefloaters/console/api/model/Node.java
```diff
@@ -4,10 +4,47 @@
...
 
 @JsonInclude(value = Include.NON_NULL)
-public record Node(int id, String host, int port, String rack) {
+public class Node {
+
+    private final int id;
+    private final String host;
+    private final int port;
+    private final String rack;
+    private Either<ReplicaInfo, Error> log;
```
tombentley, line 13: A node hosts many replicas, so why is there just a single one here? TBH it's not clear to me why it's better to model the relationship between nodes and replicas on the Node, rather than on the Partition.

api/src/main/java/com/github/eyefloaters/console/api/model/PartitionInfo.java
```diff
@@ -35,39 +36,46 @@ public PartitionInfo(int partition, Node leader, List<Node> replicas, List<Node>
...
     public void addOffset(String key, Either<OffsetInfo, Throwable> offset) {
         if (this.offsets == null) {
             this.offsets = new LinkedHashMap<>(4);
         }
 
-        this.offsets.put(key, offsetOrError(offset));
+        this.offsets.put(key, primaryOrError(offset, "Unable to fetch partition offset"));
+    }
+
+    public void addReplicaInfo(int nodeId, Either<ReplicaInfo, Throwable> log) {
```
tombentley, line 59: It seems pretty surprising that this method mutates the node. Are these public mutator methods unavoidable?

api/src/main/java/com/github/eyefloaters/console/api/model/PartitionInfo.java
```diff
@@ -90,6 +98,18 @@ public Long getRecordCount() {
             .orElse(null);
     }
 
+    public Long getSize() {
```
tombentley, line 101: We should document the semantics of this. It could be the total size over all the replicas, or the size of one replica.

api/src/main/java/com/github/eyefloaters/console/api/model/Topic.java
```diff
@@ -277,6 +279,17 @@ public Long getRecordCount() {
             .orElse(null);
     }
 
+    public BigInteger getSize() {
```
tombentley, line 282: We should document the semantics of this.

api/src/main/java/com/github/eyefloaters/console/api/service/TopicService.java
```diff
@@ -307,7 +313,7 @@ List<OffsetSpec> getRequestOffsetSpecs(String offsetSpec) {
         return specs;
     }
 
-    List<org.apache.kafka.common.TopicPartition> getTopicPartitions(Map<Uuid, Either<Topic, Throwable>> topics, Map<String, Uuid> topicIds) {
+    Map<TopicPartition, List<Integer>> getTopicPartitionNodes(Map<Uuid, Either<Topic, Throwable>> topics, Map<String, Uuid> topicIds) {
```
tombentley, line 316: Would `getTopicPartitionReplicas` be a better name?

Does it need the `get` prefix, if it's not a JavaBean getter?

api/src/main/java/com/github/eyefloaters/console/api/service/TopicService.java
```diff
@@ -319,8 +325,12 @@ List<org.apache.kafka.common.TopicPartition> getTopicPartitions(Map<Uuid, Either
                 .filter(topic -> topic.getPartitions().isPrimaryPresent())
                 .flatMap(topic -> topic.getPartitions().getPrimary()
                         .stream()
-                        .map(partition -> new org.apache.kafka.common.TopicPartition(topic.getName(), partition.getPartition())))
-                .toList();
+                        .map(partition -> {
+                            var key = new TopicPartition(topic.getName(), partition.getPartition());
+                            List<Integer> value = partition.getReplicas().stream().map(Node::getId).toList();
```
tombentley, line 330: Should partition have a `getReplicaNodeIds()` to encapsulate this?

tombentley:
I have a UX question about the screenshots... the 'Replicas' and 'In-sync Replicas' columns are counts, not node ids, right? I wonder if the 'In-sync Replicas' being a count just puts cognitive load on the user. They're basically having to compare it to the 'Replicas' column to understand what it means. It's not _such_ a problem if RF=3 uniformly, but that's not always the case:

* There are legitimate uses for topics with RF=1. 
* During a partition reassignment (one which doesn't permanently change the RF) the replicas will increase temporarily. If replicas=5 and insync=3 there's not enough context to know whether this is a problem or not.

I don't know enough about the overall UX design to really understand how problematic this might be. For example how easy it is to see if there's an ongoing reassignment that's affecting a particular partition. Or if we had plans to highlight the count of insync replicas if it was unexpectedly less than replicas.

api/src/main/java/com/github/eyefloaters/console/api/model/Node.java
```diff
@@ -4,10 +4,47 @@
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
 
 @JsonInclude(value = Include.NON_NULL)
-public record Node(int id, String host, int port, String rack) {
```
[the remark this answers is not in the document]
MikeEdgar: This changed from a record because the log field is mutable. At this point I'm questioning whether we even need to model a node like this or should it really just be the replica information with only `nodeId`. That's the spirit of what I'm attempting to do here and I think the `Node` concept is misleading. Do we ever care about the host/port/rack information?

api/src/main/java/com/github/eyefloaters/console/api/model/Node.java
```diff
@@ -4,10 +4,47 @@
...
+        this.port = port;
+        this.rack = rack;
+        this.log = null;
+    }
 
     public static Node fromKafkaModel(org.apache.kafka.common.Node node) {
         return new Node(node.id(), node.host(), node.port(), node.rack());
     }
 
+    public int getId() {
```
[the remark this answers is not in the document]
MikeEdgar: I originally left them without the `get` then found that Jackson doesn't auto-serialize them. It's nothing some annotations can't fix. I don't oppose the record-like naming for mutators/accessors and it would give us consistency between the class-based and record-based models. My preference would be to change it for all the models in a separate PR without any functional changes.

api/src/main/java/com/github/eyefloaters/console/api/model/ReplicaInfo.java
MikeEdgar: This is a bit new to me. Is that information we can get from the `Admin` interface? From the limited testing I've done with it this is node-local storage, excluding the index storage.

api/src/main/java/com/github/eyefloaters/console/api/model/Node.java
MikeEdgar: This goes back to my earlier comment about the `Node` model. Here, a `Node` represents the placement of a partition replica. Thus, the `log` is the storage used by that partition's replica on the node. As you're pointing out, this is not clear with the way it's modeled currently - or at least the way the models are named.

api/src/main/java/com/github/eyefloaters/console/api/model/Node.java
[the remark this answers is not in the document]
tombentley: host and port probably not. Rack could be useful. For example, I could imagine we want to highlight when two replicas are in the same rack, because then the redundancy is less good than it might appear.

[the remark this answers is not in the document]
tombentley: But, of course, rack being useful doesn't mean we need a `Node` class to expose it.

api/src/main/java/com/github/eyefloaters/console/api/model/PartitionInfo.java
MikeEdgar: I think the mutators will be unavoidable mainly because the main structural parts of the model are constructed earlier (e.g. using data from `describeTopics`), and the missing pieces are filled-in afterwards. This may be improved if we change how replicas are modeled.

api/src/main/java/com/github/eyefloaters/console/api/model/Node.java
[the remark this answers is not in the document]
tombentley: Sounds good. While needing annotations in the class declaration is annoying, I think consistency at call sites (of which there will be many many more) is more important.

api/src/main/java/com/github/eyefloaters/console/api/model/PartitionInfo.java
MikeEdgar: I'll add a Javadoc comment. Here, it's only the log size from the leader.

api/src/main/java/com/github/eyefloaters/console/api/model/ReplicaInfo.java
tombentley: This is a new feature known as Tiered Storage, originally described in [KIP-405](https://cwiki.apache.org/confluence/display/KAFKA/KIP-405%3A+Kafka+Tiered+Storage). Luke Chen (@showuon) would be a good person to speak to if you want more info.

For our purposes here I _think_ it's enough to adjust the docs and maybe call this property `localSize`.

api/src/main/java/com/github/eyefloaters/console/api/model/Node.java
tombentley: I did wonder about how many node instances there were floating about for a given node id! If you wanted to stick with this way of doing it then we should change the name `Node`→`ReplicaNode` or something.

api/src/main/java/com/github/eyefloaters/console/api/model/PartitionInfo.java
tombentley: Maybe be even more explicit and call it `leaderLogSize` then?

MikeEdgar:
I've seen other UIs display the node IDs for replicas/ISRs, I'm wondering if we simply display the list of node IDs for the replicas, and somehow indicate the leader (first in the list?) and then differentiate in-sync and not in-sync for the others.

For the scenario you describe with a reassignment, perhaps there are additional `Admin` operations (`listPartitionReassignments` ?) that would give enough information to give hints in the UI.

api/src/main/java/com/github/eyefloaters/console/api/model/Node.java
MikeEdgar: Suppose we do something like this:
```java
// replaces `Node` in the context of describing a replica
class Replica {
  int nodeId;
  String nodeRack;
  boolean leader;
  boolean inSync;
  // localStorage? Depends on if/how we can get tiered storage info
  Either<ReplicaStorage, Error> storage;
}

// replaces `ReplicaInfo` (too generic?)
record ReplicaStorage(
  long localSize,
  long offsetLag,
  boolean future
  ) {
}
```

The partition model could then simply have a `List` of `Replica`s, and get rid of `leader` and `isr` as separate fields.

tombentley:
> I've seen other UIs display the node IDs for replicas/ISRs, I'm wondering if we simply display the list of node IDs for the replicas, and somehow indicate the leader (first in the list?) and then differentiate in-sync and not in-sync for the others.

From Kafka's PoV `replicas` is already list and the first broker in it is the _preferred leader_. If we tried to use the order of the brokers in a list of replicas to represent the _current leader_ it would confuse people. It's also quite subtle. A specific column for the leader is impossible to misinterpret. 

I did wonder whether, rather than "in sync replicas" we showed a list of 2out of sync replicas" (i.e. just `replicas - isr`). That has the advantage that the normal happy state is always the empty list (i.e. blank) and would naturally draw the eye to those partitions where something less usual was going on. It also gives us a natural place to add extra information (e.g. a tool tip or small text explaining that we're moving a replica, or that a broker was recently restarted or whatever). 

> 
> For the scenario you describe with a reassignment, perhaps there are additional `Admin` operations (`listPartitionReassignments` ?) that would give enough information to give hints in the UI.

Yes, `listPartitionReassignments` is how you'd determine whether there was a reassignment going on. But we should only write the code once we have a clear idea of what we're aiming for.

api/src/main/java/com/github/eyefloaters/console/api/model/Node.java
tombentley: I would keep `int leaderId` in the partition model, then you could use `List<Replica> replicas`, in the same order as returned from the Admin client (thus preserving semantic of the first replica being the preferred leader). Of course you can encapsulate the lookup of the leader (on the partition model) via its id in a method so the caller doesn't have to know there isn't a `Replica leader` field.

I'd also wonder about inlining the fields of `ReplicaStorage` into `Replica` with a separate `Error storageError`. Doing it as you suggest means the structure of the objects reflects how the information was retrieved (i.e. it's only because we use a separate Admin method for getting the storage into that we have a separate class, and a indirection through an `Either`), which feels pretty invasive.  It's one thing to do that in your Java code, but when it leaks out into the JSON structure exposed by the API it starts to feel like we're exposing internal details of the implementation. At that point it might be cleaner if all the fields were nullable and each class (`Replica` etc) had a single `Error` which was non-null  whenever there was any problem. But that comes with it's own downsides for writing the code in a functional way. WDYT?

riccardo-forina:
I loved reading all the comments, there is a lot of knowledge there. But it left me wondering if we need to expose a formally correct view of how Kafka works through our APIs at all. What if we decide beforehand what is it exactly that we want to show on the console, and have the API expose just that?  It means aggregating data, simplify some concepts, put our own spices to the "Kafka console recipe". WDYT?

api/src/main/java/com/github/eyefloaters/console/api/model/Node.java
MikeEdgar: I do think there's a benefit to placing things in the `Either`, especially when it comes to a set of fields that may be inaccessible due to permissions or some combination of ACLs preventing the retrieval. Granted it does expose some info about the backend, but ultimately it does come down to the API operations that Kafka exposes, regardless of any server's implementation. 

If we later add some additional `remoteStorage` for Tiered Storage information, it compartmentalizes that storage info or the reason it is not present neatly beneath a single field in the model.

MikeEdgar:
> But it left me wondering if we need to expose a formally correct view of how Kafka works through our APIs at all. What if we decide beforehand what is it exactly that we want to show on the console, and have the API expose just that? It means aggregating data, simplify some concepts, put our own spices to the "Kafka console recipe". WDYT?

I'm looking through the model and thinking about that too. Ideally we can keep the model as DRY as possible and we'll need a way to communicate errors at a granular level + request additional data (which is supported currently with the `Either` server-side model and `union` in the Zod model).

@tombentley , I played around a bit with the replica display based on your latest comment and here's one possible view. Node 1 was stopped prior to publishing a batch of records. This shows both the out-of-sync replicas being populated as well as the leader Id being different from the preferred leader (first in list) for several partitions.

image

api/src/main/java/com/github/eyefloaters/console/api/model/Node.java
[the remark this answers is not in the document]
MikeEdgar: I reverted the `Node` changes and added `PartitionReplica` to hold a replica's `nodeId`, `nodeRack`, `inSync` status , and `localStorage`. I suspect we would have a `remoteStorage` field in the future with tiered storage information, but of course we need to wait for that to be available.

api/src/main/java/com/github/eyefloaters/console/api/model/PartitionInfo.java
MikeEdgar: Fixed this - renamed `offset` to just a generic name `either`, since `primary` would represent only one side of the `Either`.

api/src/main/java/com/github/eyefloaters/console/api/model/ReplicaInfo.java
MikeEdgar: This record was renamed to `ReplicaLocalStorage` with the `size` element kept as it was, but with a slight clarification to the description in the annotation.

api/src/main/java/com/github/eyefloaters/console/api/model/Node.java
MikeEdgar: As mentioned in the other thread, a new object `PartitionReplica` was added to carry the replica `nodeId` and `nodeRack` information, plus a boolean derived from the `isr` array. This embeds an `Either<ReplicaLocalStorage, Error>` for node-local storage of the log. I'll leave this thread open in case the discussion around how/whether to use `Either` like this should continue.

api/src/main/java/com/github/eyefloaters/console/api/model/PartitionInfo.java
MikeEdgar: The `addReplicaInfo` method and params were renamed to:
```java
setReplicaLocalStorage(int nodeId, Either<ReplicaLocalStorage, Throwable> storage)
```

api/src/main/java/com/github/eyefloaters/console/api/model/PartitionInfo.java
MikeEdgar: Renamed `getSize` on `PartitionInfo` to `leaderLocalStorage` and added a description to the API doc annotation about how it is derived.

api/src/main/java/com/github/eyefloaters/console/api/model/Topic.java
MikeEdgar: Added a description to `getSize` on the topic to indicate that it is derived as the total of all partition leader's local storage size. I also added a comment to the description that it may include remote size(s) when tiered storage is supported by the API.

api/src/main/java/com/github/eyefloaters/console/api/service/TopicService.java
MikeEdgar: Renamed `getTopicPartitionNodes` to `topicPartitionReplicas`

ui/api/topics.ts
```diff
@@ -15,21 +26,23 @@ export async function getTopic(
   kafkaId: string,
   topicId: string,
 ): Promise<Topic> {
-  const url = `${process.env.BACKEND_URL}/api/kafkas/${kafkaId}/topics/${topicId}?fields%5Btopics%5D=name,internal,partitions,authorizedOperations,configs`;
+  const url = `${process.env.BACKEND_URL}/api/kafkas/${kafkaId}/topics/${topicId}?${describeTopicsQuery}`;
   const res = await fetch(url, {
     headers: {
       Accept: "application/json",
     },
+    cache: "no-store"
```
MikeEdgar, line 34: @riccardo-forina , without this, the `fetch` only calls the server the first time from `topicPage.tsx`. Further, the storage sizes and offsets are all zero in the response that gets logged. It's not clear how that's happening or where, since the server is never called.

api/src/main/java/com/github/eyefloaters/console/api/model/Node.java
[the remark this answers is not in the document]
MikeEdgar: Opened #43 to remember to go back and use the record-style accessors throughout the model.

api/src/main/java/com/github/eyefloaters/console/api/model/Node.java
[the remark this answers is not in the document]
MikeEdgar: As noted elsewhere, reverted `Node` and added `PartitionReplica` with `nodeId` and `nodeRack`, using the record style naming :-)

ui/api/topics.ts
riccardo-forina: You did good, the default behavior of next.js is to cache all fetch responses. Adding the `cache: no-store` header opts out of this.

https://nextjs.org/docs/app/building-your-application/caching

MikeEdgar: It's surprising though that the cached value is different from what was returned from the server.

MikeEdgar:
@tombentley , @riccardo-forina please take another look and let me know if the latest addresses your concerns (or raises new ones :smiley:)

riccardo-forina (approved):
👍

tombentley:
About the [revised UI screenshot](https://github.com/eyefloaters/console/pull/42#issuecomment-1731784667) I think that's better but still not perfect.

* It's not clear that the the 'out of sync replicas' is a set of node ids. When the out of sync replicas is a singleton a user might misinterpret it as a count. (In fact in the corner case where node 0 is the one that's out of sync, the user would then think that all was well). _To me_ it would be natural to use `{` and `}` affixes to make it clear that this is a set. But I was mathematician and I'm willing to be in a minority about whether that seems natural to people more generally. We would need to do something similar for 'Replicas' to be consistent (_technically_ 'Replicas' is a list, and 'Out of sync replicas' is a set, but maybe we just treat them both as lists for the sake of the UI).  
* `0 bits` seems weird, it should be `0 bytes`

api/src/main/java/com/github/eyefloaters/console/api/model/PartitionInfo.java
```diff
@@ -26,51 +27,53 @@ private static final class OffsetInfoOrError {
...
+    public PartitionInfo(int partition, List<PartitionReplica> replicas, Integer leaderId) {
         super();
         this.partition = partition;
-        this.leader = leader;
+        this.leaderId = leaderId;
         this.replicas = replicas;
-        this.isr = isr;
     }
 
     public static PartitionInfo fromKafkaModel(org.apache.kafka.common.TopicPartitionInfo info) {
```
tombentley, line 37: Do we need the fully qualified class name, or can we import it?

api/src/main/java/com/github/eyefloaters/console/api/model/PartitionInfo.java
```diff
@@ -90,9 +93,29 @@ public Long getRecordCount() {
             .orElse(null);
     }
 
+    @JsonProperty
+    @Schema(readOnly = true, description = """
+            The total size of the log segments local to the leader replica, in bytes.
```
tombentley, line 98:
```suggestion
            The total size of the log segments local to the leader replica, in bytes.
            Or null if this is unavailable for any reason.
```

api/src/main/java/com/github/eyefloaters/console/api/model/ReplicaLocalStorage.java
```diff
@@ -0,0 +1,25 @@
+package com.github.eyefloaters.console.api.model;
+
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+public record ReplicaLocalStorage(
+        @Schema(readOnly = true, description = "The total size of the log segments local to the replica, in bytes.")
+        long size,
+
+        @Schema(readOnly = true, description = """
+                The lag of the log's LEO with respect to the partition's high watermark
```
tombentley, line 10:
```suggestion
                The lag of the log's log end offset with respect to the partition's high watermark
```

api/src/main/java/com/github/eyefloaters/console/api/model/ReplicaLocalStorage.java
```diff
@@ -0,0 +1,25 @@
...
+                (if it is the current log for the partition) or the current replica's LEO
```
tombentley, line 11:
```suggestion
                (if it is the current log for the partition), or the current replica's log end offset
```

api/src/main/java/com/github/eyefloaters/console/api/model/ReplicaLocalStorage.java
```diff
@@ -0,0 +1,25 @@
...
+                (if it is the `future` log for the partition).
+                """)
+        long offsetLag,
+
+        @Schema(readOnly = true, description = """
+                Whether this replica has been created by a AlterReplicaLogDirsRequest
+                but not yet replaced the current replica on the broker
```
tombentley, line 18: Do you think this is a better description?

```suggestion
                false this replica is the active log being used by the broker
                true if this replica represents a log being moved between log dirs on the broker  
```

Similarly I'm not totally sure `future` is the best name, even though it's the term used in the Admin API.

api/src/main/java/com/github/eyefloaters/console/api/model/Topic.java
tombentley: I wonder if we should rename it too. `Size` is ambiguous in multiple dimensions (it is bytes or records, does it include replication or not?). `TotalLeaderLogBytes` is a bit longer, but it's completely clear what it is.

Incidentally, we should consider adopting the same convention Kafka uses for quantities represented using numeric types: Always include the unit in the name of the variable, field or method. This makes it much easier when reading code and is especially helpful for times where both seconds and milliseconds are common. Messing up units is such an easy mistake to make.

api/src/main/java/com/github/eyefloaters/console/api/model/Topic.java
```diff
@@ -277,6 +279,24 @@ public Long getRecordCount() {
             .orElse(null);
     }
 
+    @Schema(readOnly = true, description = """
+            The total size, in bytes, of all log segments local to the leaders
+            for each of this topic's partition replicas.
```
tombentley, line 284:
```suggestion
            The total size, in bytes, of all log segments local to the leaders
            for each of this topic's partition replicas.
            Or null if this information is not available.
```

api/src/main/java/com/github/eyefloaters/console/api/service/TopicService.java
```diff
@@ -265,6 +270,7 @@ CompletionStage<Map<Uuid, Either<Topic, Throwable>>> describeTopics(
 
         CompletableFuture.allOf(pendingDescribes)
                 .thenCompose(nothing -> listOffsets(adminClient, result, offsetSpec))
+                .thenCompose(nothing -> describeLogDirs(adminClient, result))
                 .thenApply(nothing -> promise.complete(result))
                 .exceptionally(promise::completeExceptionally);
```
tombentley, line 275: What are we getting out of these two calls? The final `thenCompose()` already returns a `CompletableFuture`, so these two calls just seem to be completing `promise` with the same outcome as the one returned by `thenCompose()`.

api/src/main/java/com/github/eyefloaters/console/api/service/TopicService.java
```diff
@@ -307,7 +314,7 @@ List<OffsetSpec> getRequestOffsetSpecs(String offsetSpec) {
         return specs;
     }
 
-    List<org.apache.kafka.common.TopicPartition> getTopicPartitions(Map<Uuid, Either<Topic, Throwable>> topics, Map<String, Uuid> topicIds) {
+    Map<TopicPartition, List<Integer>> topicPartitionReplicas(Map<Uuid, Either<Topic, Throwable>> topics, Map<String, Uuid> topicIds) {
```
tombentley, line 317: It seems weird to me that this takes both some `topics` and also some `topicIds`. So the contract of this method is not clear from the signature alone and a sentence of Javadoc would be useful.

api/src/main/java/com/github/eyefloaters/console/api/service/TopicService.java
```diff
@@ -376,4 +391,56 @@ Either<OffsetInfo, Throwable> either(ListOffsetsResultInfo result, Throwable err
...
+
+                    return null;
+                });
+            })
+            .map(CompletionStage::toCompletableFuture)
+            .toArray(CompletableFuture[]::new);
+
+        CompletableFuture.allOf(pendingInfo)
+            .thenApply(nothing -> promise.complete(null))
+            .exceptionally(promise::completeExceptionally);
```
tombentley, line 441: This pattern again

api/src/main/java/com/github/eyefloaters/console/api/model/ReplicaLocalStorage.java
MikeEdgar: I think that description is clearer. As far as the name, let's keep it as-is for now and revisit if necessary. As you said, it's the name from the Admin API, which at least helps keep things connected.

api/src/main/java/com/github/eyefloaters/console/api/service/TopicService.java
MikeEdgar: The `thenCompose` returns are `CompletableFuture<Void>` whereas the method here returns `CompletableFuture<Map<...>>`. I find it more readable for the meaningful calls to "align" then finish up separately. Each one is enriching the `result` map, returning nothing, followed by the completion calls.

Aside - `listOffsets` and `describeLogDirs` could probably be called in parallel at some point.

current:
```java
CompletableFuture.allOf(pendingDescribes)
    .thenCompose(nothing -> listOffsets(adminClient, result, offsetSpec))
    .thenCompose(nothing -> describeLogDirs(adminClient, result))
    .thenApply(nothing -> promise.complete(result))
    .exceptionally(promise::completeExceptionally);
```

alternate:
```java
CompletableFuture.allOf(pendingDescribes)
    .thenCompose(nothing -> listOffsets(adminClient, result, offsetSpec))
    .thenCompose(nothing -> {
        describeLogDirs(adminClient, result);
        return result;
    });
```

api/src/main/java/com/github/eyefloaters/console/api/service/TopicService.java
MikeEdgar: This one can definitely be simplified :+1:

api/src/main/java/com/github/eyefloaters/console/api/service/TopicService.java
MikeEdgar: Added a comment to `topicPartitionReplicas`. The `topicIds` arg should only be temporary (hopefully) until topic ID is supported by more operations.

api/src/main/java/com/github/eyefloaters/console/api/model/Topic.java
MikeEdgar: Renamed to `TotalLeaderLogBytes` :+1:

api/src/main/java/com/github/eyefloaters/console/api/model/PartitionInfo.java
MikeEdgar: Imported TopicPartitionInfo - I think this was left from an earlier version where there was a name clash.

MikeEdgar:
> About the https://github.com/eyefloaters/console/pull/42#issuecomment-1731784667 I think that's better but still not perfect.

I agree it needs improvement. Let's use this version as the starting point for discussion and iterate from there. There will no doubt be other things to clear up the presentation as well. Unless you have any objection, I'd like to merge this one and start doing some more serious exploration around the metrics piece.

> `0 bits` seems weird, it should be `0 bytes`

Fixed in the latest, it appears at `0 B`.

image

tombentley (approved):
Thanks!

PR #44, opened 2023-09-27
Console layout changes

riccardo-forina:
"Bookmarks" is now "All resources". Also added some links on the sidebar that will eventually be part of the console.

image

The Kafka section shows a list of cards, each representing a Cluster Resource with some KPI immediately visible.
image

After selecting a Cluster, an header with some info about the cluster and an horizontal navigation is displayed. From the breadcrumb it's possible to switch to another Cluster without having to go back to the previous page.

image

image

Drilling down inside the cluster shows a similar header but with the information relative to the selected section. The breadcrumb enables going back quickly to the upper level. It's still possible to switch to another Kafka cluster from the breadcrumb: it will point to the Kafka's overview page.

image


riccardo-forina:
This depends on #42

PR #50, opened 2023-09-29
Enable UI OAuth + require OAuth-enabled brokers or no authentication

MikeEdgar:
* Refine Kafka listener selection in API layer
* Require OAuth authentication by default with option to disable for testing
* Front-end/BFF changes to support Keycloak login provider


MikeEdgar:
@tombentley , let me know if you spot any issues with the Java code. There are still some problems with the front-end dealing with the token refresh, so that can be ignored for now.

ui/app/[locale]/UserDropdown.tsx
```diff
@@ -27,16 +34,24 @@ export function UserDropdown({ username }: { username: string }) {
...
             }
           >
-            {username}
+            {username || "User"}
           </MenuToggle>
         )}
       >
         <DropdownList>
+          <DropdownItem>
+            <a href={process.env.NEXT_PUBLIC_KEYCLOAK_URL + "/account/"}>
```
Comment (MikeEdgar) [inline ui/app/[locale]/UserDropdown.tsx:50]: Can this just use `KEYCLOAK_ISSUER` ? I think `${KEYCLOAK_ISSUER} + "/account/"` is the correct path anyway.

api/src/main/java/com/github/eyefloaters/console/api/ClientFactory.java
```diff
@@ -47,11 +42,8 @@
 @ApplicationScoped
 public class ClientFactory {
```
tombentley, line 43: Can we have a sentence of javadoc explaining what the responsibilities of this class are?

api/src/main/java/com/github/eyefloaters/console/api/ClientFactory.java
```diff
@@ -132,13 +124,12 @@ Map<String, Object> buildConfiguration() {
         }
 
         Supplier<NotFoundException> noSuchKafka =
-            () -> new NotFoundException("No such Kafka cluster: " + clusterId);
+            () -> new NotFoundException("Requested Kafka cluster does not exist or is not configured with a compatible listener");
```
tombentley, line 127: Perhaps we should explain what a compatible listener is?

api/src/main/java/com/github/eyefloaters/console/api/ClientFactory.java
```diff
@@ -210,20 +169,19 @@ Map<String, Object> buildConfiguration(Kafka cluster, ListenerStatus listenerSta
...
-                })
-                .filter(credentials -> !credentials[0].isEmpty() && !credentials[1].isEmpty())
-                .map(credentials -> String.format(SASL_PLAIN_CONFIG_TEMPLATE, credentials[0], credentials[1]));
-    }
+    void configureOAuthBearer(Map<String, Object> config) {
+        log.debug("SASL/OAUTHBEARER enabled");
+        config.put(SaslConfigs.SASL_MECHANISM, OAUTHBEARER);
+        config.put(SaslConfigs.SASL_LOGIN_CALLBACK_HANDLER_CLASS, "io.strimzi.kafka.oauth.client.JaasClientOauthLoginCallbackHandler");
+        // Do not attempt token refresh ahead of expiration (ExpiringCredentialRefreshingLogin)
+        // May still cause warnings to be logged when token will expired in less than SASL_LOGIN_REFRESH_MIN_PERIOD_SECONDS.
```
tombentley, line 177:
```suggestion
        // May still cause warnings to be logged when token will expire in less than SASL_LOGIN_REFRESH_MIN_PERIOD_SECONDS.
```

api/src/main/java/com/github/eyefloaters/console/api/ClientFactory.java
```diff
@@ -210,20 +169,19 @@ Map<String, Object> buildConfiguration(Kafka cluster, ListenerStatus listenerSta
...
+        config.put(SaslConfigs.SASL_LOGIN_REFRESH_BUFFER_SECONDS, "0");
 
+        final String accessToken = Optional.ofNullable(headers.getHeaderString(HttpHeaders.AUTHORIZATION))
+                .filter(header -> header.regionMatches(true, 0, BEARER, 0, BEARER.length()))
+                .map(header -> header.substring(BEARER.length()))
+                .orElseThrow(() -> new NotAuthorizedException(BEARER.trim()));
```
tombentley, line 183: Is "Bearer" enough of an exception message to be useful?

api/src/main/java/com/github/eyefloaters/console/api/service/KafkaClusterService.java
```diff
@@ -105,33 +110,75 @@ public static Optional<Kafka> findCluster(SharedIndexInformer<Kafka> kafkaInform
                 .findFirst();
     }
 
-    public static Stream<ListenerStatus> externalListeners(Kafka kafka) {
-        return Optional.ofNullable(kafka.getStatus().getListeners())
-                .map(Collection::stream)
-                .orElseGet(Stream::empty)
-                .filter(l -> isExternalListener(kafka, l));
+    public static Optional<ListenerStatus> consoleListener(Kafka kafka) {
```
tombentley, line 113: Perhaps this should return an `Either` with the `Error` case explaining that the necessary criteria are?

api/src/main/java/com/github/eyefloaters/console/api/service/KafkaClusterService.java
```diff
@@ -105,33 +110,75 @@ public static Optional<Kafka> findCluster(SharedIndexInformer<Kafka> kafkaInform
...
+        Optional<KafkaListenerAuthentication> listenerAuth = Optional.ofNullable(listener.getAuth());
+        String authType = listenerAuth.map(auth -> auth.getType()).orElse("");
+
+        if (authType.isBlank()) {
+            return true;
+        } else if (KafkaListenerAuthenticationOAuth.TYPE_OAUTH.equals(authType)) {
+            return true;
+        } else if (KafkaListenerAuthenticationCustom.TYPE_CUSTOM.equals(authType)) {
+            return listenerAuth
+                .map(KafkaListenerAuthenticationCustom.class::cast)
```
tombentley, line 132: Can't you use an `instanceof`?

api/src/main/java/com/github/eyefloaters/console/api/service/KafkaClusterService.java
```diff
@@ -105,33 +110,75 @@ public static Optional<Kafka> findCluster(SharedIndexInformer<Kafka> kafkaInform
...
+
+    static int listenerSortKey(GenericKafkaListener listener) {
+        return annotatedListener(listener) ? -1 : 1;
+    }
+
+    static boolean annotatedListener(GenericKafkaListener listener) {
+        return Optional.ofNullable(listener.getConfiguration())
+            .map(GenericKafkaListenerConfiguration::getBootstrap)
+            .map(config -> config.getAnnotations())
+            .map(annotations -> annotations.get("eyefloaters.github.com/console-listener"))
```
tombentley, line 152: We should factor out constants for all annotations that we use. It helps with navigating the code later on. And we might not be eyefloaters for ever.

api/src/main/java/com/github/eyefloaters/console/api/ClientFactory.java
MikeEdgar: This constructor argument is a "challenge" [1] that will ultimately end up in a `WWW-Authenticate` header [2]. In the real world this code will not execute because Quarkus will have already verified that a bearer token exists and is valid per the configured JWKS/Keycloak. This aspect is disabled in the integration tests.

[1] https://jakarta.ee/specifications/restful-ws/3.1/apidocs/jakarta.ws.rs/jakarta/ws/rs/notauthorizedexception#%3Cinit%3E(java.lang.Object,java.lang.Object...)
[2] https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate

api/src/main/java/com/github/eyefloaters/console/api/service/KafkaClusterService.java
MikeEdgar: We could add a `filter(KafkaListenerAuthenticationCustom.class::isInstance)` as the first step of this chain. It could be a safety check, but at this point after checking the type is `custom` we should be safe to cast it. The only concern would be if in the future Strimzi has more than one subclass of `KafkaListenerAuthentication` that reports a `custom` type.

MikeEdgar: I think I misunderstood the line being questioned - I agree that the else/if statements would be better with `instanceof`s and I'll adjust for that.

api/src/main/java/com/github/eyefloaters/console/api/service/KafkaClusterService.java
MikeEdgar: Add an `Annotations` enum with (for now) this single value.

api/src/main/java/com/github/eyefloaters/console/api/ClientFactory.java
MikeEdgar: JavaDoc added for `ClientFactory`

MikeEdgar:
@tombentley I've addressed your comments with a few changes and mentioned some reasons why I'm thinking listener criteria is best left out from the response. Let me know if you disagree on those points.

api/src/main/java/com/github/eyefloaters/console/api/ClientFactory.java
tombentley: This sounds reasonable to me. Thanks for the detailed explanation.

api/src/main/java/com/github/eyefloaters/console/api/Annotations.java
```diff
@@ -0,0 +1,26 @@
+package com.github.eyefloaters.console.api;
+
+public enum Annotations {
+
+    /**
+     * Annotation to identify a listener in Strimzi Kafka resources to be used for
+     * connections directly from the Console API.
+     */
+    CONSOLE_LISTENER("console-listener");
```
tombentley, line 9: I guess my only other thought about this is: When do we start writing some user-centric documentation about how to deploy what we're writing. It's the sort of thing where what you need to document quickly mounts up. Right now putting something in the `README.md` would be sufficient imho.

api/src/main/java/com/github/eyefloaters/console/api/ClientFactory.java
```diff
@@ -31,30 +30,42 @@
...
-import com.github.eyefloaters.console.legacy.model.ErrorType;
 
 import io.fabric8.kubernetes.client.informers.SharedIndexInformer;
 import io.strimzi.api.kafka.model.Kafka;
 import io.strimzi.api.kafka.model.status.ListenerStatus;
 
+/**
+ * The ClientFactory is responsible for managing the life-cycles of Kafka clients
+ * - the {@linkplain Admin} client and the {@linkplain Consumer}. The factory
+ * will create a per-request client lazily when accessed by
```
tombentley, line 45:
```suggestion
 * will lazily create a per-request client when accessed by
```

api/src/main/java/com/github/eyefloaters/console/api/ClientFactory.java
```diff
@@ -31,30 +30,42 @@
...
+ * {@linkplain com.github.eyefloaters.console.api.service service code} which
+ * will be usable for the duration of the request and closed by the disposer
+ * methods in this class upon completion of the request.
```
tombentley, line 48: Just so I'm clear, this means that right if the front end needs to make multiple requests to render some UI thne we end up creating multiple clients (and multiple client authentication flows etc), right?

api/src/main/java/com/github/eyefloaters/console/api/ClientFactory.java
```diff
@@ -31,30 +30,42 @@
...
+ *
+ * <p>Construction of a client is dependent on the presence of a {@code clusterId}
+ * path parameter being present in the request URL as well as the existence of a
+ * matching Strimzi {@linkplain Kafka} CR in the watch cache available to the
+ * application's service account.
+ */
```
tombentley, line 54: Great explanation, btw. Thanks!

api/src/main/java/com/github/eyefloaters/console/api/ClientFactory.java
tombentley: Thanks, I'm missed that this was a ws exception. That's not something I'm especially familiar with, so thanks for the explanation.

api/examples/kafka-ephemeral-ingress.yaml
```diff
@@ -19,22 +19,241 @@ spec:
...
+            host: broker-2.${cluster_name}.${kube_ip}.nip.io
+      - name: oauth
+        port: 9093
+        type: ingress
+        tls: true
+        configuration:
+          bootstrap:
+            host: bootstrap-oauth.${cluster_name}.${kube_ip}.nip.io
+            annotations:
+              "eyefloaters.github.com/console-listener": "true"
```
tombentley, line 34: Anlong similar lines to the comment about doc/`README.md`, I wonder if a few comments in here would be useful to those wanting to understand this example?

api/examples/kafka-ephemeral-ingress.yaml
```diff
@@ -19,22 +19,241 @@ spec:
...
     userOperator: {}
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: ${cluster_name}-kafka-metrics
+  labels:
+    app: strimzi
+data:
+  kafka-metrics-config.yml: |
```
tombentley, line 84: If the backend has requirements for how the metrics are named let's add a comment here explaining that that is the case.

tombentley (approved):
Left a few other ideas which might be worth fixing, but overall this LGTM. Thanks @MikeEdgar!

api/src/main/java/com/github/eyefloaters/console/api/Annotations.java
MikeEdgar: > I guess my only other thought about this is: When do we start writing some user-centric documentation about how to deploy what we're writing. It's the sort of thing where what you need to document quickly mounts up. Right now putting something in the `README.md` would be sufficient imho.

Added #63 to track this.

api/examples/kafka-ephemeral-ingress.yaml
MikeEdgar: > Anlong similar lines to the comment about doc/`README.md`, I wonder if a few comments in here would be useful to those wanting to understand this example?

We can also take care of adding comments in the examples resources in #63

api/examples/kafka-ephemeral-ingress.yaml
MikeEdgar: These metrics configurations aren't actually used by the API (I had them around from some experimentation with metrics I was attempting) so there's nothing significant from a naming perspective here. I figure it's harmless enough to have in this PR and we can document metrics configuration/requirements properly once the Prometheus integration is finally included.

api/src/main/java/com/github/eyefloaters/console/api/ClientFactory.java
MikeEdgar: > Just so I'm clear, this means that right if the front end needs to make multiple requests to render some UI thne we end up creating multiple clients (and multiple client authentication flows etc), right?

That's correct. It's something we inherited from the RHOSAK design and something we'll need to keep an eye on from a scalability/performance perspective at some point.

PR #62, opened 2023-10-09
Prevent consumer seek before partition's beginning offset

### LINKED ISSUE #61: Record consumer may seek to offset before start of partition(s)
The `consumeRecords` endpoint operation may seek to an offset in one or more partitions that is before the beginning offset. This results in no records being returned for that partition.

https://github.com/eyefloaters/console/blob/1579b16280af6ad32679202df21a32d4d899cd63/api/src/main/java/com/github/eyefloaters/console/api/service/RecordService.java#L195-L206

To correct this, the beginning offsets should be fetched prior to the seek operations and the offset used should at be at minimum the beginning offset for the partition.

MikeEdgar:
Fixes #61


riccardo-forina (approved):
Thanks!

tombentley (approved):

mstruk (approved):

PR #65, opened 2023-10-10
Display any defined number value

MikeEdgar:
Several areas of the topic/partition page display dashes when there is a defined value 0.

Current:
image
From https://github.com/eyefloaters/console/issues/60#issue-1933132684

This PR displays a defined number whenever one is present, like below:
image

### DIFF
--- a/ui/components/Bytes.tsx
+++ b/ui/components/Bytes.tsx
@@ -2,5 +2,5 @@ import { useFormatBytes } from "@/utils/format";
 
 export function Bytes({ value }: { value: number | undefined }) {
   const formatter = useFormatBytes();
-  return value ? formatter(value) : "-";
+  return (typeof value !== 'undefined') ? formatter(value) : "-";
 }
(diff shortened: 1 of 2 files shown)


ui/components/Bytes.tsx
```diff
@@ -2,5 +2,5 @@ import { useFormatBytes } from "@/utils/format";
 
 export function Bytes({ value }: { value: number | undefined }) {
   const formatter = useFormatBytes();
-  return value ? formatter(value) : "-";
+  return (typeof value !== 'undefined') ? formatter(value) : "-";
```
riccardo-forina, line 5:
```suggestion
  return (value !== undefined) ? formatter(value) : "-";
```

ui/components/Number.tsx
```diff
@@ -4,5 +4,5 @@ import { useFormatter } from "next-intl";
 
 export function Number({ value }: { value: number | undefined }) {
   const formatter = useFormatter();
-  return value ? formatter.number(value) : "-";
+  return (typeof value !== 'undefined') ? formatter.number(value) : "-";
```
riccardo-forina, line 7:
```suggestion
  return (value !== undefined) ? formatter.number(value) : "-";
```

riccardo-forina (approved):
Thanks! Left a comment with an alternative check that is more familiar in js

PR #66, opened 2023-10-11
Quality of life changes

riccardo-forina:
Reordered tabs a bit, prioritising those that I feel are going the most commonly used ones by devs:
- Topics, when in a Cluster
- Messages, when in a Topic

For tabs that show a countable information I added a label with the count. This way you can enter a topic and see how many partitions it has without having to click the tab

The cluster switcher takes you to a different page depending on where you interacted with it:
- if inside one of the Cluster pages (Topics, Brokers, etc) it takes you to the same page for the new cluster
- if inside any other section, it will take you to the parent page. Eg. if you are browsing a specific topic, switching cluster will take you to the Topics page for that cluster

image
image


MikeEdgar (approved):
I haven't run it locally, but the changes look good to me from a UX perspective.

tombentley:
:+1: to the improved UX.

Issue #51, opened 2023-10-03
Broker Page


paoloantinori:

paoloantinori:
- [ ] List
- [ ] View
- [ ] Updte Configuration
- [ ] View State (disk, CPU, memory, etc)

riccardo-forina:
List of brokers

<img width="1376" alt="image" src="https://github.com/eyefloaters/console/assets/966316/6e1cb088-758e-4362-8948-c615d26abca9">

PR #67, opened 2023-10-11
Broker configuration page

riccardo-forina:
Part of #51 

Add a Broker section to the application, with a single page inside, Properties.

 In the Properties page a table shows all the properties for the selected broker, marking the property source and if it's read only. Sensitive values are masqueraded (and never reach the browser anyway).

Searching and filtering by source is implemented as well.

*Default screen, all sources are selected*
image

*Searching by property name*
image

*Filtering by property source*
image


MikeEdgar (approved):

tombentley:
:+1: to the config section.

I wonder about using the term "broker" though. KRaft brings with it a shift in terminology and a generalization. _Nodes_ are configured with a set of roles, `broker`, `controller` and `broker, controller` (a so-called _combined node_). Applications will only ever need to talk to brokers. So app devs probably only care about brokers. But Kafka cluster admins would care about controllers too. 

I think it's _probably_ OK if this section only shows information about brokers (that is, nodes whose role includes `broker`). Just be aware that if/when we think about the Kafka cluster admin persona we'd want some similar UI for controllers. If the only tab we're every going to have (for all kinds of node) is "Configuration" then we might want to just call this section "Nodes" rather than "Brokers", and provide some way to distinguish/filter based on the node roles.

riccardo-forina:
I'm ok with renaming brokers to nodes. 

It's not shown in the screenshots, but if the node is a controller it has a label next to it saying so.

We can use the same logic to describe other nodes, but I don't think that information is surfaced through the api yet (please confirm @MikeEdgar)

tombentley:
@riccardo-forina sgtm.

shirimordechay:
Consider adding a "?" tooltip with a description for nodes (Nodes are used for... nodes whose role includes broker) or include the description under the main header. If users are used to the "Broker" terminology it should be included at least in the description to provide confidence to users.

MikeEdgar:
> I'm ok with renaming brokers to nodes.
> 
> It's not shown in the screenshots, but if the node is a controller it has a label next to it saying so.
> 
> We can use the same logic to describe other nodes, but I don't think that information is surfaced through the api yet (please confirm @MikeEdgar)

I'm not aware of where to get information about the roles of the nodes, `Admin#describeMetadataQuorum` looks promising. @tombentley , would it be correct to say that `controller`s are the voters and `broker`s are the observers? Or do we need to go deeper and somehow inspect the configs for each node?

PR #68, opened 2023-10-13
Topics qol changes

riccardo-forina:
Big changes to the message browser, and to the other Topics pages.


ui/api/api.ts
```diff
@@ -0,0 +1,26 @@
...
+export async function getHeaders(): Promise<Record<string, string>> {
+  const user = await getUser();
+  return {
+    Accept: "application/json",
+    Authorization: `Bearer ${user.accessToken}`,
+  };
+}
+
+export const BackendError = z.object({
+  meta: z.object({ type: z.string() }), // z.map(z.string(), z.string()),
```
MikeEdgar, line 13: `meta` could probably be `z.any`. I didn't realize that was available when I added `BackendError` in another PR.

ui/api/topics.ts
```diff
@@ -1,28 +1,98 @@
...
-import { Topic, TopicResponse, TopicsResponse } from "@/api/types";
+import { BackendError, getHeaders } from "@/api/api";
 import { logger } from "@/utils/logger";
+import { z } from "zod";
 
 const log = logger.child({ module: "topics-api" });
 
 const listTopicsQuery = encodeURI(
-  "fields[topics]=name,internal,partitions,authorizedOperations,configs",
+  "fields[topics]=name,internal,partitions,recordCount",
```
MikeEdgar, line 8: Nice, removing `configs` saves the backend an additional API call to Kafka :+1:

ui/messages/en.json
```diff
@@ -91,23 +91,23 @@
       "offset": "Offset",
       "partition": "Partition",
       "size": "Size",
-      "timestamp": "Timestamp",
+      "timestamp": "Timestamp (local)",
+      "timestamp--utc": "Timestamp (UTC)",
```
MikeEdgar, line 96: Just pointing out the two `-` here, in case it's a typo. Other keys in the file have underscores.

MikeEdgar (approved):

ui/messages/en.json
riccardo-forina: Right, I did it like that it's because it's a variant, not a field. Translations are a mess and will have to redone anyway, I copy-pasted the old style from RHOSAK but it doesn't really work great with how this new i18n lib expects things to be.

ui/api/api.ts
riccardo-forina: Is there a unified error response from the BE?

MikeEdgar: There are some responses like for topics where there could be errors embedded - i.e. either the list of partitions or an error. This is to support situations where the user can list topics but cannot describe one or more of them. We can perhaps adjust this either/or approach if it proves unwieldy or other just ugly. It's just something I made up to support returning _something_ when something else failed during a composite operation.

Then there is a general errors response for things like validation or Kafka errors that affect the entire request. Think 404 for fetching a single topic or bad request parameters. That response follows the JSON API errors response `{ "errors": [ /* list of errors */ ] }`. The error objects are the same in either case - they match the `BackendError` zod schema.

PR #77, opened 2023-10-19
Add missing pagination links and current page number to list responses

### LINKED ISSUE #75: Add pagination `links` and current page number meta data
Pagination should include entries in the response `links` for `first`, `prev`, `next`, and `last` pages. We should also try to calculate the current page number for display purposes, accepting that it may not be accurate if records are added or deleted between requests.

MikeEdgar:
Fixes #75

### DIFF
--- a/api/src/main/java/com/github/eyefloaters/console/api/support/SizeLimitedSortedSet.java
+++ b/api/src/main/java/com/github/eyefloaters/console/api/support/SizeLimitedSortedSet.java
@@ -1,0 +1,47 @@
+package com.github.eyefloaters.console.api.support;
+
+import java.util.Comparator;
+import java.util.TreeSet;
+
+/**
+ * A SortedSet that is limited to a fixed size. Records beyond the size limit
+ * will be removed immediately. This class is useful for building paged data
+ * sets.
+ *
+ * @param <E>
+ */
+@SuppressWarnings("java:S2160")
+// Ignore Sonar warning about missing equals override, not necessary for the intended use of this class
+public class SizeLimitedSortedSet<E> extends TreeSet<E> {
+
+    private static final long serialVersionUID = 1L;
+
+    int limit;
+
+    public SizeLimitedSortedSet(Comparator<E> order, int limit) {
+        super(order);
+        this.limit = limit;
+    }
+
+    @Override
+    public boolean add(E entry) {
+        boolean added = super.add(entry);
+        truncate();
+        return added;
+    }
+
+    public int limit() {
+        return limit;
+    }
+
+    public void limit(int limit) {
+        this.limit = limit;
+        truncate();
+    }
+
+    private void truncate() {
+        while (size() > limit) {
+            pollLast();
+        }
+    }
+}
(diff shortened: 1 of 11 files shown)


MikeEdgar:
@riccardo-forina , please let me know if this doesn't match what you're expecting.

api/src/main/java/com/github/eyefloaters/console/api/support/SizeLimitedSortedSet.java
```diff
@@ -0,0 +1,47 @@
+package com.github.eyefloaters.console.api.support;
+
+import java.util.Comparator;
+import java.util.TreeSet;
+
+/**
+ * A SortedSet that is limited to a fixed size. Records beyond the size limit
+ * will be removed immediately. This class is useful for building paged data
```
tombentley, line 8: You should mention the policy by which records are removed.

api/src/test/java/com/github/eyefloaters/console/api/TopicsResourceIT.java
```diff
@@ -466,6 +471,104 @@ void testListTopicsSortedByUnknownField(String sortKey) {
             .body("data.id", contains(sortedIds));
     }
 
+    @Test
+    void testListTopicsPaginationLinksWithDefaultPageSize() throws MalformedURLException {
```
tombentley, line 475: This test would benefit from comments to explain which page you're requesting each time

api/src/test/java/com/github/eyefloaters/console/api/TopicsResourceIT.java
```diff
@@ -466,6 +471,104 @@ void testListTopicsSortedByUnknownField(String sortKey) {
...
+
+        String response1 = whenRequesting(req -> req
+                .queryParam("fields[topics]", "name")
+                .queryParam("sort", "-name")
+                .get("", clusterId1))
+            .assertThat()
+            .statusCode(is(Status.OK.getStatusCode()))
+            .body("links.size()", is(4))
+            .body("links", allOf(hasKey("first"), hasKey("prev"), hasKey("next"), hasKey("last")))
+            .body("links.first", is(nullValue()))
```
tombentley, line 497: Why is this null, rather than pointing to this same page?

api/src/main/java/com/github/eyefloaters/console/api/support/SizeLimitedSortedSet.java
MikeEdgar: Refined the JavaDoc for the `SizeLimitedSortedSet`:

https://github.com/eyefloaters/console/blob/40b13c2b2b91e91eecf1bd543b9346d55b9abe76/api/src/main/java/com/github/eyefloaters/console/api/support/SizeLimitedSortedSet.java#L7-L22

api/src/test/java/com/github/eyefloaters/console/api/TopicsResourceIT.java
MikeEdgar: Comments added to each of the page requests about the requested page and the source of the link.

tombentley (approved):
Thanks @MikeEdgar!

PR #145, opened 2023-10-31
Topics table: make the link target bigger by making it a link button

### LINKED ISSUE #135: Clickable numbers are a small target
We've made the numbers in the Kafka cluster columns into hyperlinks, which mostly works fine however when the number is a single digit (especially if it's `1`, the skinniest digit) then the clickable area is pretty small. Can we make it a bit bigger for those people who aren't super-accurate with a mouse/trackpad? 

image

riccardo-forina:
Fix #135 

image


riccardo-forina:
@tombentley the green area in the screenshot is the clickable one

MikeEdgar (approved):
:+1:

PR #149, opened 2023-11-01
Remove dependabot docker registry overrides

MikeEdgar:


riccardo-forina (approved):

MikeEdgar:
Unfortunately the dependabot.yml isn't valid to GitHub, but no clear message why and the schema doesn't appear to be published anywhere :(

Issue #53, opened 2023-10-03
Consumer Groups Page


paoloantinori:
- [ ] List
- [ ] Reset
- [ ] Offset
- [ ] View lag

PR #157, opened 2023-11-03
Consumer groups

riccardo-forina:
Depends on #110, part of #53 

* Top level consumer groups page *

image
image
image

* Topic level consumer groups page *
image


shirimordechay:
For this type of table, there should only be one expandable cell at the time.
Expandable cell: A cell that can be clicked to reveal more detail about an item. If the expansion for an item is already open, clicking on a different cell will close the current item and open a new one.
https://www.patternfly.org/components/table/design-guidelines#compound-expandable-table

PatternFly docsPatternFly docs
[PatternFly • Table](https://www.patternfly.org/components/table/design-guidelines#compound-expandable-table)
PatternFly is Red Hat's open source design system. It consists of components, documentation, and code for building enterprise applications at scale.

riccardo-forina:
> do we really need to show these details?

I don't know. It's information that Kafka provides and I think it's useful, but it would be better asked to somebody with field experience debugging consumer groups related matters. Did any requirement emerge during the interviews?

shirimordechay:
> > do we really need to show these details?
> 
> I don't know. It's information that Kafka provides and I think it's useful, but it would be better asked to somebody with field experience debugging consumer groups related matters. Did any requirement emerge during the interviews?

I agree it's useful, but isn't the same information that you would show under the "Assigned topics" details? in that case, I don't think we should show it twice. I will try to get more feedback about this area in upcoming interviews.

riccardo-forina:
I have redone the overall layout in a way that I think makes it easier to understand what's going on and how to use the consumer group pages.

image

Topics labels are links that take users to the topic's page. 
image

Clicking on a consumer group name a new page opens with all the details for all connected consumers.
image

More information can be seen expanding a row.

image

PR #169, opened 2023-11-09
First pass of the homepage

riccardo-forina:
image
image
image

@shirimordechay I have quickly put together something for the homepage. For the moment we don't have any data coming from the API to feed the "Recently changed topics"; @MikeEdgar do you think we can add the last updated timestamp from the K8s CR of the topic? I suppose that would be enough for this.

PS: the skeleton table for both topics card is because I don't yet have the data to show there, but I wanted to share this anyway to get some feedback


MikeEdgar:
> add the last updated timestamp from the K8s CR of the topic? I suppose that would be enough for this.

@riccardo-forina the main issue with that is that we aren't yet interacting with `KafkaTopic` CRs in the back-end. The topics could be created from a CR that the back-end is unaware of or those that are create via the UI function do not have an associated CR. We may need to some alternate persistence for this that is specific to the UI.

riccardo-forina:
Right! I'll simply remove the box for the moment then.

riccardo-forina:
Tracking of viewed topic implemented
image

riccardo-forina:
I also added support for the QuickStarts inline help 
image

MikeEdgar (commented):
:+1:

shirimordechay:
Looks nice! the collapsed card view will contain details as well (see example) 
image

Missing the most important values on this screen, cluster health and topics health, I assume these will be added later on?

riccardo-forina:
I don't think the Card component allows for that kind of collapsed view, with two rows of elements, [only one](https://www.patternfly.org/components/card/html-demos). You can only have the header. The collapsed view for the last box in the homepage I implemented shows that.

PR #202, opened 2023-11-20
Add filters to listTopics for `id`, `name`, and `visibility`

### LINKED ISSUE #37: Add `filter` support for topics
Allow topics to be filtered by name using a SQL-like operator or equality.

E.g.
- Fetch topic list filtered by exact name: `/topics?filter[name]=eq:topic134`
- Fetch topic list filtered by any topic with name starting with `topic`: `/topics?filter[name]=like:topic%`

MikeEdgar:
Closes #37 

Adds support for filtering topics by `id`, `name`, and `visibility`. "Internal" topics (those starting with `_`) are hidden by default any may be un-hidden by using `filter[visibility]=eq,internal` or `filter[visibility]=in,internal,external`.

Topic name filter supports `eq`, `in`, or `like` operations where `like` follows the Unix glob patterns. `*` allows any character, zero or more times; and `?` allows any single character.

### DIFF
--- a/api/src/main/java/com/github/eyefloaters/console/api/Annotations.java
+++ b/api/src/main/java/com/github/eyefloaters/console/api/Annotations.java
@@ -15,11 +15,6 @@ public enum Annotations {
         this.value = NAMESPACE + '/' + name;
     }
 
-    @Override
-    public String toString() {
-        return value();
-    }
-
     public String value() {
         return value;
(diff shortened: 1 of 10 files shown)


MikeEdgar:
Tests still needed.

riccardo-forina (approved):

Issue #165, opened 2023-11-07
Include a basic topic status 


shirimordechay:
Include topic statuses we want to surface for [MVP](https://miro.com/app/board/uXjVNbMAIcU=/?moveToWidget=3458764566856711261&cot=14):
1. In-sync- Topic is available and all partition replicas are in-sync (see under-replicated definition) 
2. Under-replicated: A partition is under-replicated if some replicas are not 'in sync'. A replica is 'in sync' if the follower has fetched records up to the log end offset within replica.lag.time.max.ms. 

The topic status will surface in the following places: 
1. Cluster overview: Show an aggregated number of in-sync/ Under-replicated topics with the ability to click and drill down to a pre-filtered topic list. 
2. Topic list: Status column, provide a tooltip when the status is under-replicated with the number of under-replicated replicas, filter list by "All statuses, In-sync, Under-replicated."

@katheris

MikeEdgar:
This is mostly completed (via #219) with the exception of tool tips and the cluster overview, which is a draft PR currently, #206.

Issue #142, opened 2023-10-30
Consistent terminology for topic configs


tombentley:
* When creating a topic we have an "Options" page which talks about "topic configuration options"
* On the topic page there is a tab for "Configuration" (which I assume will eventually show the topic config options)
* In the hamburger menu on the topics page we have "Edit properties"

These are all referring to the same conceptual thing – topic config parameters/options – so we should use consistent terminioogy

riccardo-forina:
Agree. Do we have a preference?

tombentley:
"Topic configs" would be the usual term in the AK community. But it's slightly awkward because it's an abbreviation and it's being used to refer to the many options of a single topic, where as "Topic configurations" in the normal English meaning would mean the configuration of many topics. Personally I've sometimes used "configuration parameters".

My suggestion would be:

* "configuration parameters* in the wizard
* "configuration" for the tab in the topic panel
* "edit configuration" in hamburger

shirimordechay:
I agree, Configuration fits best here and that is how it's called on the Kafka topic operator in OCP as well. 
I think "Configuration" in the wizard and topic tab and "edit configuration" from the topic kebab menu.

PR #207, opened 2023-11-22
Align the topics page to the designs

riccardo-forina:
Update the Topics page header to be in sync with the designs. The warning icon will always show zero as for the moment we don't have the count of topics that are in-sync and under replicated.

Connect the search input to search in the topic names.

Add a switch to show/hide internal topics (hide is the default). not in sync. Part of #165.

Add a status column to show the in-sync or under replicated state of a topic. This is done by looking for a replica amongst all partitions that hold the topic that is flagged as not in sync.

Reorder the columns.

Change the "Edit properties" action to "Edit configuration". Part of #142.

Change the column name "Messages" to "Record count".

Add the right empty states for when there are no topics and for when a search doesn't produce results.

image

image

image

image

### DIFF
--- a/ui/api/topics/actions.ts
+++ b/ui/api/topics/actions.ts
@@ -23,7 +23,8 @@ const log = logger.child({ module: "topics-api" });
 export async function getTopics(
   kafkaId: string,
   params: {
-    search?: string;
+    name?: string;
+    id?: string;
     pageSize?: number;
     pageCursor?: string;
     sort?: string;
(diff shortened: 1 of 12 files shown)


shirimordechay:
I don't think that record count will intuitively be clear to customers who are looking for their messages. Maybe we should keep it as "messages" and add a tooltip that explains what are we basing the numbers.

riccardo-forina:
Updated to also allow for search by topic ID

image

image

shirimordechay:
The icons for status should be: 
fa-check-circle/ fa-exclamation-triangle

riccardo-forina:
Updated icons 
image

MikeEdgar (approved):
LGTM if no other comments from @shirimordechay

shirimordechay:
We'll need to add some help tooltips to the columns, but that's probably a separate PR.
other than that LGTM.

PR #212, opened 2023-11-23
Allow selecting columns to see in the message browser

riccardo-forina:
Adds a "Manage columns" link in the Message browser toolbar
image

Clicking it will show a modal where the user can manage the visible columns
image

Any selection performed there will be persisted on the browser's local storage. Refreshing (or revisiting) the page will keep the previously selected columns. The setting is global for the message browser, it's not specific to a particular topic.

### DIFF
--- a/ui/app/[locale]/kafka/[kafkaId]/topics/[topicId]/messages/_components/ColumnsModal.tsx
+++ b/ui/app/[locale]/kafka/[kafkaId]/topics/[topicId]/messages/_components/ColumnsModal.tsx
@@ -1,4 +1,3 @@
-import { Modal } from "@/libs/patternfly/react-core";
 import {
   Button,
   DataList,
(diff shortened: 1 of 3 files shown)


shirimordechay:
+1 on managed columns. should we have one static column that is not selectable (disable the option to hide offset for example?)

MikeEdgar (commented):
One minor thing - I think it would make more sense to have the Partition column to the left of the offset column since the offset is "within" the partition.

riccardo-forina:
image

I changed the "Timestamp (locale)" to show the actual timezone in use by the browser (Europe/Rome in my case).

MikeEdgar:
> Maybe we can merge the partition and offset column, as they really have to be considered as a composite key?

I would be OK with that, either value in isolation isn't particularly helpful (in my opinion).

riccardo-forina:
Something like that? 
image
@shirimordechay WDYT?

shirimordechay:
@riccardo-forina I think this makes sense this way, especially when there is no sorting for columns. 
The row remains clickable right?

riccardo-forina:
> @riccardo-forina I think this makes sense this way, especially when there is no sorting for columns. The row remains clickable right?

Yes

PR #221, opened 2023-11-29
Multiple Kafka cluster listeners

MikeEdgar:
Return all listeners unless annotated as hidden.


MikeEdgar:
> is this based on the Access Operator?

No, the information for this comes from the Kafka custom resource.

riccardo-forina (approved):

PR #258, opened 2023-12-13
Charts are by broker, show the ranges as stacked areas

riccardo-forina:
Also finalise the layout, add proper titles and tooltips.

image

image

image
image
image


MikeEdgar (approved):
:+1:

PR #261, opened 2023-12-15
Connection details drawer

riccardo-forina:
image


riccardo-forina:
I couldn't use the QuickStarts Help feature for this, I had to write my custom drawer. Because this new drawer is nested inside the in-app help from the QuickStarts package, it was breaking the layout of the page, so I had to come up with some CSS styling to make it work. Unfortunately, this fix make it impossible to use the inline style of the drawer, so I made it an overlay.

MikeEdgar (approved):

PR #277, opened 2023-12-18
Home, cluster overview, consumer groups updates

riccardo-forina:
## Homepage

image

image

## Cluster overview (used disk space chart)

image

## Consumer groups members page

image


ui/app/[locale]/kafka/[kafkaId]/consumer-groups/[groupId]/LagTable.tsx
```diff
@@ -14,19 +14,27 @@ export function LagTable({
...
+            );
           case "partition":
             return <Th key={key}>Partition</Th>;
           case "behind":
-            return <Th key={key}>Messages behind</Th>;
-          case "offset":
-            return <Th key={key}>Offset</Th>;
+            return <Th key={key}>Lag</Th>;
+          case "currentOffset":
+            return <Th key={key}>Current offset</Th>;
```
Comment (MikeEdgar) [inline ui/app/[locale]/kafka/[kafkaId]/consumer-groups/[groupId]/LagTable.tsx:35]:
```suggestion
            return <Th key={key}>Committed offset</Th>;
```

ui/app/[locale]/home/ClustersTable.tsx
```diff
@@ -27,71 +40,105 @@ export function ClustersTable({
...
             },
             {
-              title: "Copy Bootstrap URL",
-              onClick: () => {
-                navigator.clipboard.writeText(row.attributes.bootstrapServers);
-              },
-              icon: <CopyIcon />,
+              title: "View in OpenShift console",
+              icon: <ExternalLinkAltIcon />,
+              isDisabled: true,
```
Comment (MikeEdgar) [inline ui/app/[locale]/home/ClustersTable.tsx:101]: Do we want to display the OCP link while disabled?

MikeEdgar (approved):
One question and a suggested change to the consumer group offset heading.

ui/app/[locale]/home/ClustersTable.tsx
Comment (riccardo-forina) [reply ui/app/[locale]/home/ClustersTable.tsx:101]: Can we make the link programmatically?

Comment (MikeEdgar) [reply ui/app/[locale]/home/ClustersTable.tsx:101]: We could probably read the `console` route in namespace `openshift-console` and append the path to read the Kafka CR. We should probably have an idea how we plan to handle Kubernetes (vanilla) also before we make that change though.

Comment (riccardo-forina) [reply ui/app/[locale]/home/ClustersTable.tsx:101]: Let's keep it disabled for the moment

PR #280, opened 2023-12-19
Add a way to filter topics by status

riccardo-forina:
image

image

image

### DIFF
--- a/ui/api/topics/actions.ts
+++ b/ui/api/topics/actions.ts
@@ -26,7 +26,7 @@ export async function getTopics(
   params: {
     name?: string;
     id?: string;
-    status?: TopicStatus;
+    status?: TopicStatus[];
     pageSize?: number;
     pageCursor?: string;
     sort?: string;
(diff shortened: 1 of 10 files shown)


ui/api/topics/actions.ts
```diff
@@ -38,6 +40,7 @@ export async function getTopics(
         "name,status,visibility,numPartitions,totalLeaderLogBytes,consumerGroups",
       "filter[id]": params.id ? `eq,${params.id}` : undefined,
       "filter[name]": params.name ? `like,*${params.name}*` : undefined,
+      "filter[status]": params.status,
```
MikeEdgar, line 43: In this case where there is no "operator" in the filter like `eq` or `in`, the API defaults to `eq`. If you ever want to search for multiple status values, the `in` could be used.

MikeEdgar (approved):

ui/api/topics/actions.ts
riccardo-forina: Ah it wasn't clear from the openapi doc, I thought it supported a single value only

MikeEdgar: Yeah, I missed docs for many of the `filter` parameters. They should all follow that pattern. I'll see if there's a nice way to document that without too much repetition.

riccardo-forina: Done, updated screens in the PR description to match the new option

PR #346, opened 2024-01-09
Improve the UX of the auto refresh in the Message Browser, and make it reusable

riccardo-forina:
https://github.com/eyefloaters/console/assets/966316/02f8b08a-6525-49d7-a655-e7ca27cabe6e


MikeEdgar (approved):
:+1:

PR #362, opened 2024-01-10
Add message size with tooltip, use variable polling duration

MikeEdgar:
Added "size" column with tooltip:

image

### DIFF
--- a/ui/app/[locale]/kafka/[kafkaId]/topics/[topicId]/messages/_components/MessageDetails.tsx
+++ b/ui/app/[locale]/kafka/[kafkaId]/topics/[topicId]/messages/_components/MessageDetails.tsx
@@ -23,7 +23,7 @@ import {
   TextVariants,
   Tooltip,
 } from "@/libs/patternfly/react-core";
-import { HelpIcon } from "@patternfly/react-icons";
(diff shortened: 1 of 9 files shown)


MikeEdgar:
I put the tooltip in the i18n message file, but maybe the way to extract it for the table header is not the best. @riccardo-forina , let me know if you have any suggestions on that.

ui/app/[locale]/kafka/[kafkaId]/topics/[topicId]/messages/_components/MessageDetails.tsx
```diff
@@ -20,7 +21,9 @@ import {
   Text,
   TextContent,
   TextVariants,
+  Tooltip,
 } from "@/libs/patternfly/react-core";
+import { HelpIcon } from "@patternfly/react-icons";
```
Comment (riccardo-forina) [inline ui/app/[locale]/kafka/[kafkaId]/topics/[topicId]/messages/_components/MessageDetails.tsx:26]:
```suggestion
import { HelpIcon } from "@/libs/patternfly/react-icons";
```

riccardo-forina (approved):
LGTM!

shirimordechay:
The tooltip text might change but other than that LGTM

PR #404, opened 2024-01-19
Group storybook dependabot updates

MikeEdgar:
In theory this should give us 1 storybook dependabot PR instead of 7 per day :laughing: 

Reference: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups


riccardo-forina (approved):

riccardo-forina:
Yeah I remember I did something like that for RHOSAK!

PR #422, opened 2024-01-31
Add installation resources and script for preview (non-prod) environment

MikeEdgar:


install/console-installer.sh
```diff
@@ -0,0 +1,57 @@
...
+    exit 1
+fi
+
+### Prometheus
+kubectl apply -n ${NAMESPACE} -f ${RESOURCE_PATH}/prometheus/console-prometheus-server.clusterrole.yaml
+kubectl apply -n ${NAMESPACE} -f ${RESOURCE_PATH}/prometheus/console-prometheus-server.serviceaccount.yaml
+yq '.subjects[0].namespace = strenv(NAMESPACE)' ${RESOURCE_PATH}/prometheus/console-prometheus-server.clusterrolebinding.yaml | kubectl apply -n ${NAMESPACE} -f -
+kubectl apply -n ${NAMESPACE} -f ${RESOURCE_PATH}/prometheus/kafka-resources.podmonitor.yaml
+kubectl apply -n ${NAMESPACE} -f ${RESOURCE_PATH}/prometheus/kubernetes-scrape-configs.secret.yaml
+kubectl apply -n ${NAMESPACE} -f ${RESOURCE_PATH}/prometheus/console-prometheus.prometheus.yaml
```
PaulRMellor, line 27: Is this required even if we have already deployed Prometheus?  Also, if we do already have Prometheus installed, can we just copy the `PodMonitor` and  `Secret` with the scrape configs or is it more complicated than that?

install/resources/kafka/cluster-example.kafka.yaml
```diff
@@ -0,0 +1,65 @@
...
+      inter.broker.protocol.version: '3.6'
+      min.insync.replicas: 2
+      offsets.topic.replication.factor: 3
+      transaction.state.log.min.isr: 2
+      transaction.state.log.replication.factor: 3
+    listeners:
+      - name: secure
+        port: 9093
+        tls: true
+        type: ${LISTENER_TYPE}
```
PaulRMellor, line 24: We can use any listener type supported by AMQ Streams?  We don't support ingress downstream, only on Strimzi. Route is expected downstream.

install/resources/kafka/cluster-example-metrics.configmap.yaml
```diff
@@ -0,0 +1,183 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: cluster-example-metrics
```
PaulRMellor, line 4: I wonder if we should use _examples_ in the names? I'm assuming this should not be changed.

MikeEdgar: I may not be following, do you mean it should be called `cluster-examples-metrics`?

install/install-community-operators.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
MikeEdgar: This one will install the community versions of Strimzi and Prometheus if the user hasn't already installed them. I'm a little unsure how to deal with this one because a tech preview user likely won't be using the community Strimzi in practice.

install/resources/kafka/cluster-example-metrics.configmap.yaml
PaulRMellor: Not clear! I was thinking something like `cluster-metrics`, as I'm assuming these metrics configurations shouldn't be adjusted. Using _example_ in the name sounds like this is only one way of configuring the metrics

install/install-community-operators.sh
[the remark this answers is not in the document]
PaulRMellor: More assumptions on my part: assuming we'd be deploying (and referring in the docs) to streams operators downstream. The way I've written the connection procedure is that you use this script to deploy AMQ Streams or configure your current deployment of AMQ Streams. The doc can bend to any realities for the tech preview...

install/resources/kafka/cluster-example-metrics.configmap.yaml
MikeEdgar: I agree. The naming of this and other resources in the same directory are all like `cluster-example` to align with the name of the Kafka CR. Maybe we can change them all to `demo-cluster` or `trial-cluster`? I'm fine with other suggestions too.

install/install-community-operators.sh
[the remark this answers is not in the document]
MikeEdgar: What do you think about calling it `install-dependency-operators.sh` ? Of all the scripts, only `console-install.sh` is required, with the others being optional depending on the current state of the user's cluster.

install/resources/kafka/cluster-example-metrics.configmap.yaml
PaulRMellor: If we uses `demo-cluster-metrics`, I wonder if we're in danger of making the name obsolete when it moves beyond demo?  The only other name I can think of is `console-kafka-metrics`, as we're scarping kafka metrics for the console.

MikeEdgar: I think the Kafka resources will always be for demonstration purposes to some extent. Let's go with `console-kafka` rather than `demo-cluster` for these resources.

MikeEdgar:
@PaulRMellor , @jankalinic if no objects I'd like to merge this.

jankalinic (approved):

PR #469, opened 2024-02-12
Configure Kafka connections at startup, OAuth or anonymous UI access

MikeEdgar:
1. User "login"
  image
2. Account menu after logging in
  image
3. Read only, no buttons to create topic, no menus to edit/delete topics
  image
4. Topic configuration edit disabled in read only mode
  image


riccardo-forina:
@shirimordechay, Mike did the changes we discussed about Monday and instead of removing completely the authentication he kept it with a fake flow, that doesn't require any username and password. What do you think about it UX wise?

shirimordechay:
I think that we could probably add some description about the anonymous user restrictions, either in the login screen or in some overlay that shows up when login in as anonymous.

riccardo-forina (approved):
Tested locally, thank you for the changes

riccardo-forina:
PR looks good to me. @shirimordechay I'd let @MikeEdgar merge this so we can update the demo environment. In the meanwhile if you want to suggest changes in terms of messages or other feedback for the user, we can do that in a second moment.

MikeEdgar:
Should we add some alert/warning banner across the top when using anonymous sessions? It can be in a subsequent PR probably.

PR #470, opened 2024-02-13
Add page level playwright tests

suyash-naithani:


ui/jest.setup.ts
```diff
@@ -0,0 +1,27 @@
...
+let page: Page;
+
+const username = process.env.APP_USERNAME ?? "";
+const password = process.env.APP_PASSWORD ?? "";
+
+beforeAll(async () => {
+  browser = await chromium.launch();
+  page = await browser.newPage();
+  await page.goto(
+    "https://console.amq-streams-ui.us-east.containers.appdomain.cloud/home"
```
riccardo-forina, line 13: this should be coming from an env variable as well, end everywhere where the domain is used should use the env variable as well

ui/tests/playwright/ClusterOverview.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+      "Key performance indicators and important information regarding the Kafka cluster."
+    );
+    const noMessagesText = await page.$eval('span[id="no-messages"]', (span) =>
+      span?.textContent?.trim()
+    );
+    expect(noMessagesText).toBe("No messages");
+    //const page = await page.innerText('div[class="pf-v5-l-flex pf-m-justify-content-center pf-m-wrap pf-m-nowrap-on-sm pf-m-gap"]');
+    expect(await page.innerText("body")).toContain("kafka1");
+    expect(await page.innerText("body")).toContain("Ready");
+    expect(await page.innerText("body")).toContain("3/3");
```
riccardo-forina, line 19: You can't really have tests for the values without controlling the environment. These values will be invalid the moment the cluster is used for something, like for the demo data generation, that will change the amount of disk space used/available.

It's enough to check for the labels for the moment.

riccardo-forina (commented):
I left a couple comments.

jankalinic (commented):
I agree with Riccardo's notes. When testing environment cannot be controlled to insert certain values, then checking them would not be a good idea.

ui/tests/playwright/BrokerPropertyPage.test.tsx
```diff
@@ -0,0 +1,11 @@
+import { page } from "../../jest.setup";
+describe("Brokers property page", () => {
+  test("Brokers property page should display table", async () => {
+    await page.goto(
+      "https://console.amq-streams-ui.us-east.containers.appdomain.cloud/kafka/j7W3TRG7SsWCBXHjz2hfrg/nodes/0/configuration"
```
MikeEdgar, line 5: Can we look at running the application in minikube or something similar rather than referencing the demo environment? Otherwise, how can these tests verify the contents of this repo and the changes made in PRs, etc?

suyash-naithani: This will not verify if the changes in the PR will break the demo env, onlv checks if the demo env is presently running as expected. I will start this discussion on slack to weigh our options on this

ui/tests/playwright/BrokerPropertyPage.test.tsx
```diff
@@ -6,7 +6,7 @@ describe("Brokers property page", () => {
     await page.goto(
       `${URL}/kafka/j7W3TRG7SsWCBXHjz2hfrg/nodes/0/configuration`
     );
-    await page.waitForLoadState("networkidle");
+    await page.waitForLoadState("networkidle",{timeout: 60000});
```
riccardo-forina, line 9: Why not setting it as the default timeout instead of overriding each wait? https://playwright.dev/docs/test-timeouts

Also, why is this needed? The default of 30s should be plenty.

ui/playwright.config.ts
```diff
@@ -0,0 +1,16 @@
+import { defineConfig } from 'playwright/test';
+
+export default defineConfig({
+  // Run your local dev server before starting the tests
+  webServer: {
+    command: 'npm run dev',
+    url: 'https://console-api-eyefloaters-dev.mycluster-us-east-107719-da779aef12eee96bd4161f4e402b70ec-0000.us-east.containers.appdomain.cloud',
```
riccardo-forina, line 7: This url isn't right, you want to test against localhost:3000, that's the whole purpose of running the dev server with Playwright

ui/playwright.config.ts
```diff
@@ -0,0 +1,16 @@
...
+    reuseExistingServer: !process.env.CI,
+    stdout: 'ignore',
+    stderr: 'pipe',
+    timeout: 120 * 1000,
+  },
+  use: {
+    baseURL: 'https://console-api-eyefloaters-dev.mycluster-us-east-107719-da779aef12eee96bd4161f4e402b70ec-0000.us-east.containers.appdomain.cloud',
```
riccardo-forina, line 14: Same as above

riccardo-forina (changes requested):

riccardo-forina (changes requested):
I see you used deprecated APIs like the $ and $$ selectors (doc [here](https://playwright.dev/docs/api/class-elementhandle#element-handle-query-selector)).

Considering that this is new code, could you please use locators as by suggested by the doc?

PR #488, opened 2024-02-16
Improve scrolling for the message browser and properties pages

riccardo-forina:
No padding
image
Drawer using all the space
image
Scrolling of the table without breaking the page/drawer
image

Same for the properties
image

This removes the padding around the tables to give more space to the content, and works around the problem with drawers breaking the inner scrolling of the table and drawer by specifying a fixed height to the table.


riccardo-forina:
@shirimordechay wdyt?

riccardo-forina:
Closing in favor of #500

PR #500, opened 2024-02-20
Message browser improvements

riccardo-forina:


shirimordechay:
@riccardo-forina I think the manage columns action could be displayed as an icon(+hover tooltip with action name) 
_fa-columns_, since the kebab icon only contains that single option we can simply replace it. 
"I have found that people tend to get confused by the dual selection panels and avoid using them. Is there a particular reason why you chose to use this component? I believe a simple list, with the ability to drag and reorder items, would be easier to understand."

riccardo-forina:
Thanks @shirimordechay, I changed the column management as suggested:

image


image

MikeEdgar (approved):

PR #504, opened 2024-02-21
Initial support for managed KafkaTopic resources

MikeEdgar:
Added a "Managed" indicator to the topic status in the listing, just to demonstrate the data is available. A topic is managed when there is a corresponding `KafkaTopic` CR in Kubernetes.

image

Also added to subtitle for topic details, but I couldn't figure out how to add an icon...
image

### DIFF
--- a/api/.checkstyle/checkstyle.xml
+++ b/api/.checkstyle/checkstyle.xml
@@ -97,7 +97,7 @@
 
         <module name="ClassFanOutComplexity">
             <!-- default is 20 -->
-            <property name="max" value="44"/>
+            <property name="max" value="50"/>
             <property name="excludedPackages" value="java.time,javax.json,javax.ws.rs,javax.ws.rs.core"/>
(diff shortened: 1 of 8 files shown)


MikeEdgar:
@shirimordechay @riccardo-forina please take a look and let me know your feedback. I'm sure the UX will need some work :)

shirimordechay:
> @shirimordechay @riccardo-forina please take a look and let me know your feedback. I'm sure the UX will need some work :)

I think it doesn't make sense to add that information to the status column, it probably would make more sense to add it as a compact label under/ side to the name column example: 
image
)
We can place the same label on the topic page under /the side of the name. 

I would also add a tooltip on hover for that label to clarify what "Managed" stands for, as I'm sure it will raise some questions.

riccardo-forina:
image
image

riccardo-forina (approved):

PR #514, opened 2024-02-27
[install] MacOS modify creation of console secrets

jankalinic:
This pull request addresses an issue where the `tr` command encountering an "Illegal byte sequence" error when attempting to generate a random string from `/dev/urandom`. The issue is typically related to locale settings when using MacOS.


MikeEdgar (approved):
Thanks for the PR @jankalinic . I appreciate the testing on MacOS too!

PR #527, opened 2024-03-01
Add more Storybook stories

riccardo-forina:
This is a first pass at improving the test coverage of the UI. I added stories with tests for all the components in the components directory, and restored the ones for the components we migrated over from RHOSAK


MikeEdgar (approved):
Should we also add the `test-storybook` script to the CI?

riccardo-forina:
It's not strictly necessary as we have the Chromatic action that, while building the Storybook, runs the tests as well. Failures are reported as a failed action

PR #544, opened 2024-03-04
fix: identify configured and un-configured Kafka clusters

### LINKED ISSUE #543: Kafka clusters in returned list do not indicate whether a connection is configured
Kafka clusters in returned list do not indicate whether configuration is present. The result is that the front-end has no way to know if a cluster may be linked/navigable in the UI.

The API should indicate in the cluster list whether each cluster is configured or not, using the `meta` section of each `data` entry. The UI can then display plain/non-link text in the cluster list on the home screen, and exclude the cluster from the navigation tree.

MikeEdgar:
Fixes #543 

Changes:
 - links disabled for Kafka clusters without connection configuration
 - broker/consumer groups show _Connection not configured_ rather than details
 - Clusters excluded from the navigation

Sample
image

### DIFF
--- a/ui/app/[locale]/home/page.tsx
+++ b/ui/app/[locale]/home/page.tsx
@@ -97,7 +97,7 @@ export default function Home() {
                     </Tooltip>
                   </b>
                   <Text component={"small"}>
-                  The last 5 topics this account has accessed from the AMQ 
+                  The last 5 topics this account has accessed from the AMQ
                   Streams console.
                   </Text>
                 </TextContent>
@@ -297,7 +297,14 @@ async function ConnectedClustersTable({
   const allClusters = await clusterPromise;
   const clusters = allClusters.map<EnrichedClusterList>((c) => {
     async function getNodesCounts() {
-      const cluster = await getKafkaCluster(c.id);
+      let cluster;
+
+      if (c.meta.configured === true) {
+        cluster = await getKafkaCluster(c.id);
+      } else {
+        cluster = null;
+      }
+
       if (cluster) {
         return {
           count: cluster.attributes.nodes.length,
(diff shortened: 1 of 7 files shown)


ui/app/[locale]/home/page.tsx
```diff
@@ -297,7 +297,14 @@ async function ConnectedClustersTable({
...
   const clusters = allClusters.map<EnrichedClusterList>((c) => {
     async function getNodesCounts() {
-      const cluster = await getKafkaCluster(c.id);
+      let cluster;
+
+      if (c.meta.configured === true) {
+        cluster = await getKafkaCluster(c.id);
+      } else {
+        cluster = null;
+      }
```
Comment (riccardo-forina) [inline ui/app/[locale]/home/page.tsx:306]: I think we want this in the getCluster, so to protect the pages that depend on it and properly redirect users to the homepage

riccardo-forina (approved):
Thanks! I only had a comment

ui/app/[locale]/home/page.tsx
Comment (MikeEdgar) [reply ui/app/[locale]/home/page.tsx:306]: I see what you mean. Since it's called from the homepage itself, where does the redirect happen?

Comment (riccardo-forina) [reply ui/app/[locale]/home/page.tsx:306]: It doesn't, it's handled on a per-route basis checking for a null return value.

Comment (MikeEdgar) [reply ui/app/[locale]/home/page.tsx:306]: In that case, maybe getKafkaCluster just needs to handle the 404 error. It won't have access to the meta information from the cluster list entry (without an additional request). Wdyt?

Comment (MikeEdgar) [reply ui/app/[locale]/home/page.tsx:306]: Returning null on 404 from `getKafkaCluster` results in a page error, mostly from other calls for consumer groups, etc. failing. I think the only way to get to those pages would be to obtain the cluster Id indirectly and use it in the URL, correct?

Comment (riccardo-forina) [reply ui/app/[locale]/home/page.tsx:306]: Right my comment was to cover the case of users typing the url by hand, or getting one shared, or from a bookmark

Comment (riccardo-forina) [reply ui/app/[locale]/home/page.tsx:306]: It's fine to keep it this way for the moment

shirimordechay:
It would be helpful to include a help icon with a link to more information about how to configure or an actionable item of some kind along with the "Connection not configured" text.

riccardo-forina:
Agree. Could you please open an issue with the details about how you suggest doing it?

PR #555, opened 2024-03-06
fix: avoid replacing Admin client when Kafka CR is missing certificates

MikeEdgar:


api/src/main/resources/application.properties
```diff
@@ -59,8 +59,6 @@ quarkus.arc.unremovable-types=com.github.eyefloaters.console.api.**
 
 kafka.admin.request.timeout.ms=10000
 kafka.admin.default.api.timeout.ms=10000
-kafka.admin.sasl.login.refresh.min.period.seconds=60
-kafka.admin.sasl.login.refresh.buffer.seconds=60
```
MikeEdgar, line 63: These apply to OAuth2, which isn't always used anymore. Removing them to prevent the Kafka client from logging warnings about their use.

riccardo-forina (approved):

PR #637, opened 2024-04-03
Configurable build args

riccardo-forina:


MikeEdgar (approved):

riccardo-forina:
I'm actually not sure this is needed. But it should do no harm either. WDYT?

MikeEdgar:
Right, I don't see any harm. As it is, the default values should be the same if the args aren't given at build time.

PR #734, opened 2024-05-06
Multi-platform images

MikeEdgar:


MikeEdgar:
@riccardo-forina you're expecting `linux/arm64` images for running on MacOS right?

riccardo-forina (approved):

riccardo-forina:
Yes thank you!

riccardo-forina:
Will this tag the console-api image as `latest` as well as the ui one? By the way, wouldn't it make sense to have the ui image be called `console-ui` for consistency with the api one?

MikeEdgar:
> Will this tag the console-api image as `latest` as well as the ui one? By the way, wouldn't it make sense to have the ui image be called `console-ui` for consistency with the api one?

It's so far just building the images in the test CI and not yet in the release job. I agree we should use `console-ui` for the image name. I'll make that change here as well. It would also be good to add a new job that builds the images on each commit to `main` and push a `main` tag into quay.io.

MikeEdgar:
@riccardo-forina , the ARM build seems to be quite slow. The output of `npm run build` isn't architecture-specific is it? If it is not, maybe the build could be re-factored somehow.

PR #807, opened 2024-06-05
Rebrand project StreamsHub

MikeEdgar:


MikeEdgar:
@riccardo-forina , I'm getting this error when the topics list is empty. Any guidance you can offer is appreciated. The intent is to globally enable/disable learning links and also allow some of them to be unset by default and later overridden via the messages file.
image

install/000-install-dependency-operators.sh
```diff
@@ -12,7 +12,7 @@ echo -e "${INFO} Create/update operator group in namespace '${NAMESPACE}'"
 ${YQ} '.spec.targetNamespaces[0] = strenv(NAMESPACE)' ${RESOURCE_PATH}/prometheus/console-operators.operatorgroup.yaml | ${KUBE} apply -n ${NAMESPACE} -f -
 
 if ${KUBE} get packagemanifests amq-streams >/dev/null ; then
```
paoloantinori, line 14: probably not a big deal, but it's still `amq-streams` in the command

MikeEdgar: The name of the package manifest (and other k8s resources) hasn't changed yet.

PR #809, opened 2024-06-07
Use KRaft for example install, support node pool storage metrics

MikeEdgar:
With Strimzi KafkaNodePools [1] (required to run the cluster in KRaft mode), storage metrics have a slightly different format for the PVC label. This PR adds support for storage allocated via a node pool, and also changes the example Kafka cluster to use KRaft mode.

[1] https://strimzi.io/blog/2023/08/28/kafka-node-pools-storage-and-scheduling/


riccardo-forina (approved):
If this backward compatible?

MikeEdgar:
> If this backward compatible?

Yes, it's meant to display the storage information for either Zookeeper-based or KRaft-based clusters. Of course the script change for the example install now only uses KRaft. We can keep the ZK variant as well - wdyt?

MikeEdgar:
I've added back a way to run the Kafka cluster with ZK (additional script argument) and this is mentioned now in the `install/README.md`

PR #828, opened 2024-06-17
[WIP] - [CI] Add minikube deployment for PR testing

jankalinic:
In this PR a new GitHub action workflow is created. This workflow is triggered on every pull-request which targets main branch (for each commit).

It utilizes installation of strimzi operator, prometheus operator from online example. 
Prometheus instance, kafka cluster and console deployment are installed using scripts from examples in this repository (located in `./install/ directory`). 

After an environment is prepared, console images are built and console is deployed. We should discuss what this step could be, if images can be built using make or can depend on existing build workflow. There is a demonstrative curl step at the end of the workflow which should be replaced with some testing steps.

In order to fully make this work a `console-config.yaml` was added based on the other provided examples style.
Also a few changes i the installation scripts needed to be done (Like removing unused variable `PROVIDED_APIS`) and switch that skips operatorhub installation on ci/minikube cluster.

Note: For demonstration purposes deployment images needed to be changed. After this PR is reviewed and approved these images will be replaced with streamshub ones (they do not exist at this point in the streamshub registry).

### DIFF
--- a/.github/workflows/minikube-deployment.yaml
+++ b/.github/workflows/minikube-deployment.yaml
@@ -1,0 +1,155 @@
+name: Minikube Deployment
+
+on:
+  workflow_call:
+
+env:
+  TARGET_NAMESPACE: "console-namespace"
+  CI_CLUSTER: true
+  CONFIG_YAML: "./install/resources/console/console-config.yaml"
+  DEPLOYMENT_YAML: "./install/resources/console/console.deployment.yaml"
+  CLUSTER_DOMAIN: "192.168.49.2.nip.io"
+  CONSOLE_API_IMAGE: "quay.io/streamshub/console-api:latest"
+  CONSOLE_UI_IMAGE: "quay.io/streamshub/console-ui:latest"
+
+jobs:
+  setup_deployment:
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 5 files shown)


.github/workflows/pr-job.yaml
```diff
@@ -0,0 +1,102 @@
...
+      - name: Wait for Kafka
+        run: kubectl wait kafka/console-kafka --for=condition=Ready --timeout=300s -n $TARGET_NAMESPACE
+
+      - name: Show pods
+        run: kubectl get pods -n $TARGET_NAMESPACE
+
+      - name: Wait for user secret
+        run: while ! kubectl get secret console-kafka-user1 -n $TARGET_NAMESPACE; do echo "Waiting for user secret"; sleep 5; done
+      # TODO: requires images on streamshub - just like in examples
+#     Build images and run deployment
```
jankalinic, line 70: This step is here as a placeholder, we should build images here for the later usage.

.github/workflows/pr-job.yaml
```diff
@@ -0,0 +1,102 @@
...
+      - name: Show pods
+        run: kubectl get pods -n $TARGET_NAMESPACE
+
+      - name: Add Hosts Entry
+        shell: bash
+        run: |
+          echo "${MINIKUBE_IP} console-ui.${CLUSTER_DOMAIN}" | sudo tee -a /etc/hosts
+
+      # TODO: remove once satisfied with results
+      - name: Test Curl
```
jankalinic, line 99: This step is also a placeholder for some testing.

PR #839, opened 2024-06-21
Initial implementation of console operator

MikeEdgar:
Add initial operator project based on java-operator-sdk using dependent resources [1]. This PR also includes a new Github workflow to publish snapshot images upon each PR merge to main.

Done:
- [x] Basic functionality to deploy the console in its current state
- [x] Operator basic testing
- [x] Build/deploy OLM resources
- [x] Script to install operator, documentation to create Console CR

Upcoming PRs:
- Fine-tune RBAC needed by operator (and console REST API itself)
- Informers/event handlers for `Kafka` and `KafkaUser`/`Secret` modifications
- Split Kafka cluster reference out from `Console` CR to new sub-resource
- Support for multiple Prometheus install modes (e.g. using Prom operator, direct, none)

[1] https://javaoperatorsdk.io/docs/dependent-resources/


MikeEdgar:
@k-wall , @katheris, FYI if you have a chance to look through this PR and spot any issues.

operator/src/main/java/com/github/streamshub/console/ConsoleReconciler.java
```diff
@@ -0,0 +1,170 @@
...
+        if (!readyStatus.equals(readyCondition.getStatus())) {
+            readyCondition.setStatus(readyStatus);
+            readyCondition.setLastTransitionTime(Instant.now().toString());
+        }
+
+        if (isReady) {
+            readyCondition.setReason(null);
+            readyCondition.setMessage("All resources ready");
+        } else {
+            readyCondition.setReason("DependentsNotReady");
```
k-wall, line 153: I've just learnt another UK/US spelling variation :)

MikeEdgar: Are you referring to dependent vs dependant? IMO this isn't the right spelling even for US in my experience.

MikeEdgar: But I just copied it to match the method `getNotReadyDependents`

operator/src/main/java/com/github/streamshub/console/dependents/ConsoleResource.java
```diff
@@ -0,0 +1,101 @@
...
+            .sorted(Comparator.comparing(Map.Entry::getKey))
+            .forEach(entry -> digest.update(entry.getValue().getBytes(StandardCharsets.UTF_8)));
+
+        setAttribute(context, digestName, digest);
+    }
+
+    default String serializeDigest(Context<Console> context, String digestName) {
+        var resourceContext = context.managedDependentResourceContext();
+        return resourceContext.get(digestName, MessageDigest.class)
+                .map(digest -> String.format("%040x", new BigInteger(1, digest.digest())))
```
k-wall, line 90: Java 17 HexFormat instead?

MikeEdgar: I think it's older than 17. I copied this approach from what was done in kas-fleetshard to trigger rolling of the admin server when secrets were changed.

https://github.com/bf2fc6cc711aee1a0c2a/kas-fleetshard/blob/d2122a64dafb6db6aba90fa14002de077dc300ff/operator/src/main/java/org/bf2/operator/managers/SecuritySecretManager.java#L157-L177

MikeEdgar: > Java 17 HexFormat instead?

I misread your comment, I'll look into `HexFormat`

operator/src/main/resources/application.properties
```diff
@@ -0,0 +1,20 @@
+console.selector=
+
+quarkus.container-image.build=true
+#quarkus.container-image.group=
+quarkus.container-image.name=console-operator
+
+quarkus.operator-sdk.activate-leader-election-for-profiles=prod
+quarkus.operator-sdk.controllers."consolereconciler".selector=${console.selector}
```
k-wall, line 8: why the quotes?  hang over from the old name?

operator/src/test/example-console.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: console.streamshub.github.com/v1alpha1
+kind: Console
+metadata:
+  name: example
+  namespace: streams-console
+spec:
+  hostname: example-console.apps-crc.testing
+  kafkaClusters:
+    - kafkaUserName: console-kafka-user1
```
k-wall, line 9: does this refer to the name of a kafkauser resource?

operator/pom.xml
```diff
@@ -0,0 +1,231 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
k-wall, line 1: Could you add information for a developer who is trying to spin this up locally?

operator/src/test/example-console.yaml
k-wall: I see the answer to that is yes.  I don't think the field name conveys that well.

MikeEdgar:
I've updated the install README [1] and removed the separate Prometheus deployment from the scripts. That is handled by the operator now and we can revisit in the future when we add support for the Prometheus operator.

In short, you should be able to get things running by these steps. This assumes Strimzi and the Console are in separate namespaces, so set the `STRIMZI_NAMESPACE` and `CONSOLE_NAMESPACE` envs accordingly. The `CLUSTER_DOMAIN` env can be set to the domain of your OpenShift/K8s cluster.
1. `./install/000-install-strimzi-operator.sh ${STRIMZI_NAMESPACE}`
1. `./install/001-install-console-operator.sh ${CONSOLE_NAMESPACE}`
1. `./install/002-deploy-console-kafka.sh ${STRIMZI_NAMESPACE} ${CLUSTER_DOMAIN}`
1. ```shell
   echo 'apiVersion: console.streamshub.github.com/v1alpha1
   kind: Console
   metadata:
     name: example
   spec:
     hostname: example-console.'${CLUSTER_DOMAIN}' # Hostname where the console will be accessed via HTTPS
     kafkaClusters:
       - name: console-kafka                       # Name of the `Kafka` CR representing the cluster
         namespace: '${STRIMZI_NAMESPACE}'         # Namespace of the `Kafka` CR representing the cluster
         listener: secure                          # Listener on the `Kafka` CR to connect from the console
         kafkaUserName: console-kafka-user1        # Name of the `KafkaUser` resource used to connect to Kafka
   ' | kubectl apply -n ${CONSOLE_NAMESPACE} -f -
   ```

When complete, you should have a console at `https://example-console.${CLUSTER_DOMAIN}`

[1] https://github.com/streamshub/console/blob/operator/install/README.md

operator/src/test/example-console.yaml
MikeEdgar: I've been thinking about the name of this field a bit too and I'm not too happy with the name either. I haven't been able to come up with anything yet, do you have any ideas that make more sense?

operator/src/main/resources/application.properties
MikeEdgar: Usually, whenever there is some application-provided element of a property, it is quoted in the full property key. It's likely because the name of the reconciler could have a dot character itself and this clarifies that the quoted element is a variable section. This is the case for log configuration also, where the FQCN (the log category) is a single element of the property key, but itself contains dots.

operator/pom.xml
MikeEdgar: I will add something about this. If you get to looking at it before I do, you can try the following command from the repo root to start the operator in dev mode using your current Kube context. Assuming you have Strimzi and a `Kafka` CR in the Kube cluster, you can create a `Console` CR like in my other comment [1].

```
mvn -am -pl operator quarkus:dev
```

[1] https://github.com/streamshub/console/pull/839#issuecomment-2195612729

operator/src/test/example-console.yaml
riccardo-forina: what about `kafkaUser_username`? It combines the kind and the field we need, something that can be easily retrieved running `kubectl get kafkauser -A -o jsonpath="{.items[*].status.username}"`

MikeEdgar: This actually represents the `.metadata.name` of the `KafkaUser`. It's used to fetch the name of the generated secret from the status of the `KafkaUser`.

riccardo-forina: `kafkaUser.metadata.name` then. Or `kafkaUser_metadata_name`. It might be verbose, but it's clear. And can be obtained easily `kubectl get kafkauser -A -o jsonpath="{.items[*].metadata.name}"  `

operator/src/main/java/com/github/streamshub/console/api/v1alpha1/spec/ConsoleSpec.java
```diff
@@ -0,0 +1,55 @@
...
+public class ConsoleSpec {
+
+    String hostname;
+
+    Images images = new Images();
+
+    List<KafkaCluster> kafkaClusters = new ArrayList<>();
+
+    // TODO: copy EnvVar into console's API to avoid unexpected changes
+    List<EnvVar> env = new ArrayList<>();
```
MikeEdgar, line 22: I left this TODO here because I thought it might be bad form to embed something from fabric8 directly in the CRD, but I see there is precedent for that in the Strimzi API as well. Unless anyone has reservations about doing this, maybe we just keep it as-is until there's a compelling reason to duplicate `EnvVar` into this project.

operator/pom.xml
```diff
@@ -0,0 +1,231 @@
...
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>io.quarkiverse.operatorsdk</groupId>
+            <artifactId>quarkus-operator-sdk</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.fabric8</groupId>
+            <artifactId>crd-generator-apt</artifactId>
```
metacosm, line 57: You shouldn't need this dependency

operator/pom.xml
```diff
@@ -0,0 +1,231 @@
...
+        <container-image.registry>localhost</container-image.registry>
+        <container-image.group>streamshub</container-image.group>
+        <container-image.tag>${project.version}</container-image.tag>
+        <container-image.push>false</container-image.push>
+    </properties>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>io.quarkus</groupId>
```
metacosm, line 28: Any particular reason you're using the "internal" BOM as opposed to the platform one?

metacosm: For some reference: https://developers.redhat.com/articles/2023/09/19/write-operators-java-josdk-part-4-upgrading-strategies#strategies_for_qosdk_and_quarkus_updates

MikeEdgar: I've updated this and the plugin references to use `io.quarkus.platform` using properties. Thanks

operator/src/main/java/com/github/streamshub/console/dependents/ConsoleResource.java
MikeEdgar: This was replaced with `HexFormat` which should be more clear. Thanks for the suggestion.

operator/pom.xml
MikeEdgar: Thanks. Replaced `crd-generator-apt` with fabric8:generator-annotations and sundrio:builder-annotations.

metacosm: Hmm maybe I missed something: why were you using this dependency? QOSDK should generate the CRD for you automatically without needing to run the annotation processor…

operator/src/main/java/com/github/streamshub/console/api/v1alpha1/spec/ConsoleSpec.java
```diff
@@ -0,0 +1,55 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+import io.fabric8.kubernetes.api.model.EnvVar;
+import io.sundr.builder.annotations.Buildable;
+
+@Buildable(builderPackage = "io.fabric8.kubernetes.api.builder")
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class ConsoleSpec {
+
+    String hostname;
```
katheris, line 15: Are any of these fields required?

operator/src/test/example-console.yaml
katheris: If you're using Strimzi it's possible to have the KafkaUser CR in a different namespace to the Kafka CR, so what about:
```
kafkaUser:
   name: console-kafka-user1
   namespace: streams-console (optional)
```

In the Strimzi Kafka Access operator we actually use:
```
user:
    kind: KafkaUser
    apiGroup: kafka.strimzi.io
    name: my-user
    namespace: kafka
```
With the idea that in future we could support different types of user, we don't have a clear idea of what those would be, but just to make it easily extensible

operator/src/main/java/com/github/streamshub/console/dependents/ConsoleSecret.java
```diff
@@ -0,0 +1,218 @@
...
+        String name = kafkaRef.getName();
+        String listenerName = kafkaRef.getListener();
+
+        KafkaClusterConfig kcConfig = new KafkaClusterConfig();
+        kcConfig.setNamespace(namespace);
+        kcConfig.setName(name);
+        kcConfig.setListener(listenerName);
+        config.getKafka().getClusters().add(kcConfig);
+
+        // TODO: add informer for Kafka CRs
```
katheris, line 127: Does this mean at present if the Kafka CR is changed, e.g. the listener address changes the console will not react to the change? Or is Java operator SDK running a timed reconciliation as well as reacting to resource updates?

operator/src/test/example-console.yaml
MikeEdgar: > If you're using Strimzi it's possible to have the KafkaUser CR in a different namespace to the Kafka CR, so what about:
> 
> ```
> kafkaUser:
>    name: console-kafka-user1
>    namespace: streams-console (optional)
> ```
> 
> In the Strimzi Kafka Access operator we actually use:
> 
> ```
> user:
>     kind: KafkaUser
>     apiGroup: kafka.strimzi.io
>     name: my-user
>     namespace: kafka
> ```
> 
> With the idea that in future we could support different types of user, we don't have a clear idea of what those would be, but just to make it easily extensible

I did not realize that. In that case, I liked what you've done in the access operator and we should probably do something like that here.

operator/src/main/java/com/github/streamshub/console/dependents/ConsoleSecret.java
MikeEdgar: It currently means it will not react to the change, but the SDK does have a timed reconciliation of 10 hours (!). I'll push an update to lower that to 60 seconds or so. Unfortunately, it doesn't seem to be an interval that can be configured at runtime, so I'll also open an issue if there isn't already one.

api/pom.xml
```diff
@@ -37,9 +36,9 @@
                 <version>9.40</version>
             </dependency>
             <dependency>
-                <groupId>io.quarkus</groupId>
-                <artifactId>quarkus-bom</artifactId>
-                <version>${quarkus.version}</version>
+                <groupId>${quarkus.platform.group-id}</groupId>
```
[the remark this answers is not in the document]
MikeEdgar: This is something that shows up in many current example Quarkus applications, and it's because there is a set of platforms [1] that could be selected based on the application's use case and possibly swapped. I may be missing something on that however - perhaps @metacosm may have a better understanding on the reason for using properties.

[1] https://github.com/quarkusio/quarkus-platform

MikeEdgar:
> The main thing I was struggling to follow was how the install works. Do the shell scripts generate the Console CRDs on the fly? Would it be worth doing that in advance and checking them into version control? It ought to be possible to have the install just be a set of .yaml files I think.

The build is actually generating the CRDs and packaging them into an OLM bundle, which is then added to a single-bundle OLM catalog. The install script `001-install-console-operator.sh` adds a resource for the catalog source and a subscription.

MikeEdgar:
> If you're using Strimzi it's possible to have the KafkaUser CR in a different namespace to the Kafka CR

@katheris can you confirm the above? I thought the `KafkaUser` must be in the same namespace as the `Kafka`, and are only operated on by the entity operator created for that `Kafka` CR.

operator/src/main/java/com/github/streamshub/console/ConsoleReconciler.java
```diff
@@ -0,0 +1,201 @@
+package com.github.streamshub.console;
```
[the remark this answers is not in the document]
MikeEdgar: > Do the Java classes need an Apache 2.0 license header?

My understanding is that having a `LICENSE` in the project root is sufficient from a legal perspective.

operator/src/main/java/com/github/streamshub/console/dependents/ConsoleSecret.java
MikeEdgar: For reference: https://github.com/quarkiverse/quarkus-operator-sdk/issues/893

operator/src/main/java/com/github/streamshub/console/api/v1alpha1/spec/ConsoleSpec.java
MikeEdgar: Only `hostname` is required. It's been annotated as such in the latest commit.

operator/pom.xml
MikeEdgar: > Hmm maybe I missed something: why were you using this dependency? QOSDK should generate the CRD for you automatically without needing to run the annotation processor…

I must have picked it up from an example or older blog post, then came to rely on it for the transitive deps fabric8:generator-annotations and sundrio:builder-annotations, which are now direct.

operator/src/main/java/com/github/streamshub/console/dependents/ConsoleSecret.java
MikeEdgar: Added a periodic reconciliation interval of 60s. Let me know if that doesn't seem like a reasonable value.

operator/src/test/example-console.yaml
MikeEdgar: (moved comment to thread)

> If you're using Strimzi it's possible to have the KafkaUser CR in a different namespace to the Kafka CR

@katheris can you confirm the above? I thought the `KafkaUser` must be in the same namespace as the `Kafka`, and are only operated on by the entity operator created for that `Kafka` CR.

katheris: You specify an alternative namespace, see the docs https://strimzi.io/docs/operators/latest/deploying#deploying-the-user-operator-using-the-cluster-operator-str
```
You configure the entityOperator property of the Kafka resource to include the userOperator. By default, the User Operator watches for KafkaUser resources in the namespace of the Kafka cluster deployment. You can also specify a namespace using watchedNamespace in the User Operator spec. A single User Operator can watch a single namespace. One namespace should be watched by only one User Operator.
```

katheris:
> The build is actually generating the CRDs and packaging them into an OLM bundle, which is then added to a single-bundle OLM catalog. The install script `001-install-console-operator.sh` adds a resource for the catalog source and a subscription.

So is OLM a pre-req for installing the console operator?

MikeEdgar:
> So is OLM a pre-req for installing the console operator?

At the moment, yes. However, the operator can be run locally and the CRDs are generated & deployed to Kube for dev mode.

operator/src/test/example-console.yaml
MikeEdgar: Thanks, that helps. How about something like this where we can use a polymorphic `user` with a `type`. Initially we could support a `KafkaUser`, but it could be extended to other things like `Secret` later. wdyt?

```yaml
user:
  type: KafkaUser
  name: console-kafka-user1
  namespace: streams-console (optional)  
```

```yaml
user:
  type: Secret
  name: console-kafka-user1
  namespace: streams-console (optional)
  # both required
  usernameKey: <...>
  passwordKey: <...>
  # or alternatively use jaas config
  jaasConfigKey: <...>
```

katheris: Yeah that looks reasonable to me

MikeEdgar: It turns out that support for polymorphism in the tooling is not very good. I ended up going with this:

```yaml
credentials:
  # docs: Reference to a Strimzi KafkaUser resource
  user:
    name: console-kafka-user1
    namespace: streams-console (optional)
```

The expectation is that this could be expanded in the future to this, where exactly one of `user` or `secret` must be provided.

```yaml
credentials:
  # docs: Reference to a Strimzi KafkaUser resource
  user:
    name: console-kafka-user1
    namespace: streams-console (optional)
  secret:
    name: console-kafka-user1
    namespace: streams-console (optional)
    # both required
    usernameKey: <...>
    passwordKey: <...>
    # or alternatively use jaas config
    jaasConfigKey: <...>
```

MikeEdgar:
@metacosm , @katheris , @k-wall , @riccardo-forina please take another look if you can. I'd like to merge this early next week. We can make gradual improvements going forward for at least those things listed in the first comment above.

README.md
```diff
@@ -27,7 +30,7 @@ Please refer to the [installation README](./install/README.md) file for detailed
 ### Run locally
 
 Running the console locally requires the use of a remote or locally-running Kubernetes cluster that hosts the Strimzi Kafka operator
-and any Apache Kafka® clusters that will be accessed from the console. To get started, you will need to provide a console configuration
+and any Apache Kafka<sup>®</sup> clusters that will be accessed from the console. To get started, you will need to provide a console configuration
```
k-wall, line 33: (non-blocking) I'd provide the minimal step required to deploy Strimzi in a suitable way.  We want the bar for get started with a contribution to be as low as we can.

k-wall (approved):

k-wall:
It was quite a big PR to absorb, but looks reasonable.  Well done @MikeEdgar.

As we discussed elsewhere, we should spend time refactoring the READMEs so the developer getting started experience is as smooth as possible. This is an enabler for getting more folks contributing and reviewing.

PR #864, opened 2024-07-02
Build UI direct in action, simplify UI Dockerfile, fix API release args

MikeEdgar:
Also correct reference to environments repo


MikeEdgar:
@riccardo-forina please take a look. This should get the build performance back to what it was prior to the multi-arch images being added. We can do another release once this is merged.

riccardo-forina (approved):

PR #902, opened 2024-07-15
[OLM] Fix operator UI and add new fields for operator

jankalinic:
This PR fixes errors with operator display page/sidebar in openshift UI. Also it adds new fields such as creation timestamp, support, certified, categories, repository link, operator image sha etc. The procedure of creating a catalog image with such fields is following -> First a CSV yaml file is created from CSVMetadata annotation using operator-sdk. Then it's placeholder values are replaced using yq and skopeo (both need to be installed) and a catalog image is created.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -60,8 +60,7 @@ jobs:
             -Dquarkus.container-image.push=true
 
       - name: Modify CSV Annotation
-        run: |
-          ./operator/bin/modify-bundle-metadata.sh
+        run: ./operator/bin/modify-bundle-metadata.sh
 
       - name: Build Operator Bundle Image
         uses: docker/build-push-action@v6
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -9,6 +9,11 @@ on:
 jobs:
   build-api:
     runs-on: ubuntu-latest
+    services:
+      registry:
+        image: registry:2
+        ports:
+          - 5000:5000
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -32,6 +37,8 @@ jobs:
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
+        with:
+          driver-opts: network=host
 
       - name: Build and Test
         env:
@@ -48,7 +55,13 @@ jobs:
           #
           mvn verify -P container-image -B --no-transfer-progress \
             -Dquarkus.package.write-transformed-bytecode-to-build-output=true \
-            -Dquarkus.docker.buildx.platform=linux/amd64,linux/arm64
+            -Dquarkus.docker.buildx.platform=linux/amd64,linux/arm64 \
+            -Dquarkus.container-image.registry="localhost:5000" \
+            -Dquarkus.container-image.push=true
+
+      - name: Modify CSV Annotation
+        run: |
+          ./operator/bin/modify-bundle-metadata.sh
 
       - name: Build Operator Bundle Image
         uses: docker/build-push-action@v6
@@ -68,7 +81,7 @@ jobs:
           sudo cp -v opm /usr/bin/
           rm -vf opm
           VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout | tr '[:upper:]' '[:lower:]')
-          operator/bin/generate-catalog.sh ${VERSION} 
+          operator/bin/generate-catalog.sh ${VERSION}
 
       - name: Build Operator Catalog Image
         uses: docker/build-push-action@v6
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -12,6 +12,11 @@ jobs:
     runs-on: ubuntu-latest
     name: snapshot
     if: ${{github.event.pull_request.merged == true}}
+    services:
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -25,9 +29,9 @@ container-image-api-push: container-image-api
 
 container-image-operator:
 	mvn package -am -pl operator -Pcontainer-image -DskipTests -Dquarkus.container-image.image=$(CONSOLE_OPERATOR_IMAGE)
-	operator/bin/generate-catalog.sh $(VERSION)
-	$(CONTAINER_RUNTIME) build -t $(CONSOLE_OPERATOR_BUNDLE_IMAGE) -f operator/target/bundle/console-operator/bundle.Dockerfile
-	$(CONTAINER_RUNTIME) build -t $(CONSOLE_OPERATOR_CATALOG_IMAGE) -f operator/target/catalog.Dockerfile
+	operator/bin/generate-catalog.sh VERSION=$(VERSION) ARCH=$(ARCH) CONSOLE_OPERATOR_IMAGE=$(CONSOLE_OPERATOR_IMAGE) SKIP_RANGE=$(SKIP_RANGE)
```
MikeEdgar, line 32: The `integration` github action doesn't use the Makefile, but executes `generate-catalog.sh` directly. I think that's the reason for the CI failure.

jankalinic: Thanks, i missed that.

jankalinic:
I think this time it's not related to this PR. it fails on PW test

```
// Click on the "Connection details" button within the dropdown menu
    > 23 |     await page.click(
         |                ^
      24 |       'button[class="pf-v5-c-menu__item"] span[class="pf-v5-c-menu__item-text"]:has-text("Connection details")'
      25 |     );
      ```

File: operator/bin/generate-catalog.sh [hunk 2]
```diff
@@ -1,16 +1,64 @@
...
+
+    curr_time_date=$(date +'%Y-%m-%d %H:%M:%S')
+    echo "[DEBUG] Setting createdAt = ${curr_time_date}"
+    ${YQ} eval -o yaml -i ".metadata.annotations.createdAt = \"${curr_time_date}\"" "${CSV_FILE_PATH}"
+
+    if [[ -n "$SKIP_RANGE" ]]; then
+        echo "[DEBUG] Setting skipRange = \"${SKIP_RANGE}\""
+        ${YQ} eval -o yaml -i ".metadata.annotations.[\"olm.skipRange\"] = \"${SKIP_RANGE}\"" "${CSV_FILE_PATH}"
+    fi
+}
```
MikeEdgar, line 51: It just occurred to me that this is modifying the bundle in the script that is for building the catalog. Maybe there should be a separate job step after the build/test step and before `Build Operator Bundle Image` that does this update. This script and the catalog build step wouldn't change then (I think).

.github/workflows/integration.yml
```diff
@@ -46,10 +53,20 @@ jobs:
...
             -Dquarkus.package.write-transformed-bytecode-to-build-output=true \
             -Dquarkus.docker.buildx.platform=linux/amd64,linux/arm64
 
+      - name: Modify CSV Annotation
+        run: |
+          docker ps
+          docker images 
+          ls ./operator/target/bundle/console-operator/manifests/
+          cat ./operator/target/bundle/console-operator/manifests/console-operator.clusterserviceversion.yaml
+          ./operator/bin/modify-csv.sh
```
MikeEdgar, line 68: Old name? Should it be `modify-bundle-metadata.sh`?

operator/bin/modify-bundle-metadata.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+SCRIPT_PATH="$(cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P)"
+source ${SCRIPT_PATH}/common.sh
+
+ARCH=linux/amd64
```
MikeEdgar, line 6: Can `ARCH` be removed?

.github/workflows/integration.yml
```diff
@@ -9,6 +9,11 @@ on:
 jobs:
```
MikeEdgar, line 9: Do you want to make the equivalent changes to `.github/workflows/snapshot.yml` too? That will build and put an image to quay.io once this merges to main and we can verify that. If all looks good, a follow-up PR to do `.github/workflows/release.yml` will be the last piece.

operator/bin/modify-bundle-metadata.sh
jankalinic: Sorry missed that one, thanks.

.github/workflows/snapshot.yml
```diff
@@ -12,6 +12,11 @@ jobs:
     runs-on: ubuntu-latest
     name: snapshot
     if: ${{github.event.pull_request.merged == true}}
+    services:
+      registry:
+        image: registry:2
+        ports:
+          - 5000:5000
```
MikeEdgar, line 19: Ah, sorry. Since snapshot already pushes to quay, we shouldn't need the local registry or the `driver-opts` on the `setup-buildx-action`.

MikeEdgar: I think later we can perhaps use the local registry and then copy them all at once if everything works. Not perfect, but slightly more "atomic" than what we have now.

PR #913, opened 2024-07-17
[Operator] Add more csv fields to improve operator display in Openshift UI

jankalinic:
This PR is a follow-up which adds more CSV metadata fields. That should improve overall display in Openshift UI and help with search using streams and kafka related keywords.

### DIFF
--- a/operator/src/main/java/com/github/streamshub/console/ConsoleReconciler.java
+++ b/operator/src/main/java/com/github/streamshub/console/ConsoleReconciler.java
@@ -43,6 +43,7 @@ import io.quarkiverse.operatorsdk.annotations.CSVMetadata;
 import io.quarkiverse.operatorsdk.annotations.CSVMetadata.Annotations;
 import io.quarkiverse.operatorsdk.annotations.CSVMetadata.Annotations.Annotation;
 import io.quarkiverse.operatorsdk.annotations.CSVMetadata.InstallMode;
+import io.quarkiverse.operatorsdk.annotations.CSVMetadata.Link;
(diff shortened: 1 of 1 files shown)


jankalinic:
image

operator/src/main/java/com/github/streamshub/console/ConsoleReconciler.java
```diff
@@ -133,12 +134,35 @@
                 @Annotation(name = "createdAt", value = "placeholder"),
                 @Annotation(name = "support", value = "Streamshub")
             }),
-        description = "StreamsHub console provides a user interface for managing and monitoring your streaming resources",
+        description = "The Streamshub Console provides a web-based user interface tool for monitoring Apache Kafka® instances within a Kubernetes based cluster.\n\n "
```
MikeEdgar, line 137: This might be better as a text block since it's multiple lines. E.g.
```java
description = """
   The StreamsHub Console provides...
   """
```
You can omit the `\n` characters in that case because newlines are taken literally in the block. wdyt?

MikeEdgar (commented):
Very nice, just one suggestion to make the description easier to read/maintain in the code.

operator/src/main/java/com/github/streamshub/console/ConsoleReconciler.java
jankalinic: I did this at first, but it resulted in black background and code-style-like looking font, this way it looks clean. But i don't like that its like this either. It makes the editing more complicated.

jankalinic:
NOTE: Needed to remove my fork, therefore the PR is closed. Will open a new PR following this one.

PR #914, opened 2024-07-17
[CI] Fix and unify image tags to lowercase, fix cut on localhost

jankalinic:
This PR fixes some minor issues where images were being tagged with uppercase, while catalog was created with lower case image tag. This resulted in images missing in registry and back-off errors during operator deployment.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,17 +20,17 @@ jobs:
         with:
           token: ${{secrets.RELEASE_TOKEN}}
 
-      - name: Set Image Tag Env
-        run: |
-          echo "PROJECT_VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout | tr '[:upper:]' '[:lower:]')" >> $GITHUB_ENV
-          echo "CURRENT_VERSION=$(echo ${{steps.metadata.outputs.current-version}} | tr '[:upper:]' '[:lower:]' )" >> $GITHUB_ENV
-
       - name: Retrieve Project Metadata
         uses: radcortez/project-metadata-action@603060a0627d4f04d37d20331ebd472377fd088b
         id: metadata
         with:
           github-token: ${{secrets.GITHUB_TOKEN}}
           metadata-file-path: '.github/project.yml'
+
+      - name: Set Image Tag Env
+        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -53,6 +58,7 @@ jobs:
           export QUARKUS_CONTAINER_IMAGE_PASSWORD="${{ secrets.IMAGE_REPO_PASSWORD }}"
           export QUARKUS_CONTAINER_IMAGE_PUSH="true"
           export QUARKUS_CONTAINER_IMAGE_ADDITIONAL_TAGS=latest
+          export QUARKUS_CONTAINER_IMAGE_TAG="${{ env.PROJECT_VERSION }}"
```
MikeEdgar, line 61: I think exporting here for the release workflow won't work because the POM file still has the `-SNAPSHOT` version, but running the `release:perform` will build the non-SNAPSHOT. However, I think it is not an issue because there is nothing to actually change for the non-SNAPSHOT, so perhaps the release job doesn't need to change at all (as long as we stick with x.y.z releases).

.github/workflows/snapshot.yml
```diff
@@ -23,6 +23,11 @@ jobs:
           metadata-file-path: '.github/project.yml'
           local-file: 'true'
 
+      - name: Set Image Tag Env
+        run: |
+          echo "PROJECT_VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout | tr '[:upper:]' '[:lower:]')" >> $GITHUB_ENV
+          echo "NEXT_VERSION=$(echo ${{steps.metadata.outputs.next-version}} | tr '[:upper:]' '[:lower:]')" >> $GITHUB_ENV
```
MikeEdgar, line 29: These should both be the same value, maybe using just `NEXT_VERSION` shows more clearly that the snapshot is for "next".

MikeEdgar (changes requested):
Looks good. Thanks @jankalinic . A few comments and I think the release job may not need to change since it only uses numeric anyway.

.github/workflows/release.yml
jankalinic: In that case i'll remove the changes. Thanks

MikeEdgar (approved):
LGTM :+1:

PR #916, opened 2024-07-18
[API] Obtain credentials from request when missing configuration

MikeEdgar:


api/src/main/java/com/github/streamshub/console/api/ClientFactory.java
```diff
@@ -522,6 +562,42 @@ Map<String, Object> buildConfig(Set<String> configNames,
...
+        switch (authType) {
+            case "oauth":
+                saslEnabled = true;
+                cfg.putIfAbsent(SaslConfigs.SASL_MECHANISM, OAUTHBEARER);
+                cfg.putIfAbsent(SaslConfigs.SASL_LOGIN_CALLBACK_HANDLER_CLASS, STRIMZI_OAUTH_CALLBACK);
+                break;
+            case "scram-sha-512":
+                cfg.putIfAbsent(SaslConfigs.SASL_MECHANISM, SCRAM_SHA512);
+                saslEnabled = true;
+                break;
```
riccardo-forina, line 580: Shouldn't there be a case for SCRAM SHA 256 as well?

MikeEdgar: Strimzi doesn't support 256, so this switch will only be 512. The configuration could have specified 256 however, so I can add that below where the user/pass are taken from the request header.

riccardo-forina (commented):
I'm still testing this against the UI changes. The SCRAM login works fine, I'm working on the oauth one.

So far so good anyway

riccardo-forina (approved):

PR #923, opened 2024-07-22
[Operator] Add Console Instance Description

jankalinic:
This PR adds CSV metadata to Console Instance card displayed in openshift UI, which currently has "Not available" description.

Signed-off-by: jkalinic <jkalinic@redhat.com>

### DIFF
--- a/operator/src/main/java/com/github/streamshub/console/api/v1alpha1/Console.java
+++ b/operator/src/main/java/com/github/streamshub/console/api/v1alpha1/Console.java
@@ -9,12 +9,17 @@ import io.fabric8.kubernetes.api.model.Namespaced;
 import io.fabric8.kubernetes.client.CustomResource;
 import io.fabric8.kubernetes.model.annotation.Group;
 import io.fabric8.kubernetes.model.annotation.Version;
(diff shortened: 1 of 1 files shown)


operator/src/main/java/com/github/streamshub/console/api/v1alpha1/Console.java
```diff
@@ -9,12 +9,17 @@
...
 import io.sundr.builder.annotations.Buildable;
 
 @Version("v1alpha1")
 @Group("console.streamshub.github.com")
 @Buildable(editableEnabled = false, builderPackage = "io.fabric8.kubernetes.api.builder")
 @JsonInclude(JsonInclude.Include.NON_NULL)
+@CSVMetadata(
+    displayName = "Console Instance",
+    description = "Represents a deployable Console UI for monitoring Kafka resources."
+    )
```
MikeEdgar, line 22: Does this really work for CRD metadata? I thought the CRDs were generated by Fabric8. I'm trying to understand in the JOSDK code how this works.

MikeEdgar (approved):
Confirmed that this does indeed work - the information is placed in the CSV and not in the CRD as I incorrectly assumed. Thanks for the PR @jankalinic

PR #941, opened 2024-07-31
Single cluster support, with dynamic authentication

riccardo-forina:
# Changes

## Homepage

Doesn't require authentication, shows the available clusters.

image

## Login screen for clusters with SCRAM-SHA authN

image

## Login screen for clusters with OAuth (token) authN

image

## Login screen for unauthenticated clusters

image

## Post-login homepage (cluster overview), and all other screens

Side-navigation shows only the cluster's links. The recently viewed topic is here now. The rest of the app isn't changed.
To log in to a different cluster, users will have to logout first; they'll land on the homepage.

image


MikeEdgar:
The brokers page doesn't work for me and I'm having a tough time debugging it. I pushed a commit with the middleware pattern fix.

riccardo-forina:
Thank you. What error are you getting in the brokers page? I tested it without a Prometheus connection set up, and the page was showing ok with the warning that there was missing data. I didn't test it again against a working Prometheus

MikeEdgar:
The nodes table does not display at all, but the header and Prometheus warning do display. I am also not connected to a Prometheus instance. I've tried setting log statements where the API call to the Kafka cluster is made, but they aren't being shown on the console, even though the API server logs show requests are being made :sob: 

Generally, I also can improve things by deleting the `.next` directory and restarting, which makes me think the caching is again causing problems. Note, that did not help with the nodes page.

riccardo-forina:
Yeah without Prometheus you will not see Anthony at all other than the warning and the table header. It's not great, but I'm not sure what else we can show there without Prometheus. But I'll check again tomorrow

MikeEdgar (approved):
Verified working with https://github.com/streamshub/console/pull/916

shirimordechay:
- I don't see value in keeping the collapsable card option for the clusters login list. we can use a simple card
- Does the Tech Preview badge appear in the upstream version as well? 
- Should the Prometheus warning only show once instead of repeatedly displaying the message?

MikeEdgar:
> * I don't see value in keeping the collapsable card option for the clusters login list. we can use a simple card
> 
> * Does the Tech Preview badge appear in the upstream version as well?
> 
> * Should the Prometheus warning only show once instead of repeatedly displaying the message?

@shirimordechay , the Tech Preview badge is present upstream. I need to defer to @riccardo-forina (when he returns) and @hemahg for the other two questions :smile:

hemahg:
@shirimordechay, yes we can make use of a simple card. collapsable card is not needed in this case.
For the Prometheus warning, using the <AlertActionCloseButton> is a good approach to allow users to dismiss the alert when they see it. wdyt

shirimordechay:
I think providing an error with a link to documentation on how to enable Prometheus would be helpful for users who are unsure how to proceed. I'm not sure what you meant by "allowing users to dismiss the alert when they see it." Is the alert dismissible?

hemahg:
Currently, the alert as implemented is not dismissible by default. My previous suggestion to include an <AlertActionCloseButton> would make the alert dismissible, allowing users to close it manually if they see it.
However, considering your suggestion, providing an error message with a link to documentation on how to enable Prometheus could indeed be very helpful.

PR #955, opened 2024-08-06
Add API/UI health probes, enable operator health probes

MikeEdgar:


ui/app/healthz/route.ts
```diff
@@ -0,0 +1,6 @@
+export async function GET() {
+  return Response.json({
+    "status": "UP",
+    "checks": [],
+  });
```
MikeEdgar, line 5: This format matches the health responses from the API and operator.

riccardo-forina: Does it have to live in the root? Can we move it in the api folder?

riccardo-forina (approved):
Looks good

ui/app/healthz/route.ts
MikeEdgar: It's more of a convention that I see, but it could be anywhere I think. Looking at this link [1], I see that `healthz` is a deprecated convention and it's better to use `livez` and `readyz`. I guess we can just keep `healthz` since there is no distinction between the two states for the UI (for now).

[1] https://kubernetes.io/docs/reference/using-api/health-checks/

PR #957, opened 2024-08-07
Allow operand images to be set via configuration properties (or env)

MikeEdgar:


operator/src/main/resources/application.properties
```diff
@@ -1,5 +1,9 @@
 console.selector=
 
+console.deployment.default-image-tag=${quarkus.application.version}
+console.deployment.default-api-image=quay.io/streamshub/console-api:${console.deployment.default-image-tag}
+console.deployment.default-ui-image=quay.io/streamshub/console-ui:${console.deployment.default-image-tag}
```
MikeEdgar, line 5: These properties impact the images that are applied to the console's `Deployment` resource when none are given via the `Console` CR's `spec.images.api` or `spec.images.ui` properties. 

Any thoughts on the naming in particular? I don't expect these would typically be specified very often aside from perhaps pinning a specific SHA in the operator's ClusterServiceVersion.

cc: @k-wall

k-wall:
Just checking I understand, so in the CSV for the operator we'll define the env vars for `CONSOLE_DEPLOYMENT...IMAGE` so that the bundle will control the default image used for API and UI.   The build pipeline will take care of making sure the right value is injected.

If the user wishes to override the image, he may do so by overriding within the CR.   He's also got the option of overriding the env vars in the Subscription.

lgtm

k-wall (approved):

jankalinic (approved):

MikeEdgar:
> Just checking I understand, so in the CSV for the operator we'll define the env vars for `CONSOLE_DEPLOYMENT...IMAGE` so that the bundle will control the default image used for API and UI. The build pipeline will take care of making sure the right value is injected.
> 
> If the user wishes to override the image, he may do so by overriding within the CR. He's also got the option of overriding the env vars in the Subscription.

Yes, that's correct.

PR #976, opened 2024-08-20
operator: obtain default image tag from label

MikeEdgar:
- obtain default image tag from label
- skip all previous versions for snapshot and CI builds
- add host name to info log message
- add namespace variable to generate K8s resources
- attach generate K8s resource to builds and releases


MikeEdgar:
@jankalinic please take a look. This change accounts for the snapshot image tags being lower-case by telling the operator to use the lower-case name rather than its built-in default with upper-case `-SNAPSHOT`. 

This also adds an "everything" skipRange on the snapshot builds to make them easily testable. We can/should think about how to calculate skipRange on the release builds too.

MikeEdgar:
I am going ahead with merging these build changes. I would like to verify the behavior in the release process and make any corrections if necessary.

PR #1024, opened 2024-09-05
Implement consumer group patch (reset offsets) "dry run" option

MikeEdgar:
Implement consumer group patch (reset offsets) "dry run" option required by #960


MikeEdgar:
@hemahg , this should give you what is needed for dry-run in #960 . Simply pass `"dryRun": true` in a `"meta"` consumer group patch object. You will receive a response with the array of `offsets`, just like is returned from the consumer group `GET` operation.

hemahg:
Thanks @MikeEdgar, I will implement the dry-run page on top of this PR

hemahg (approved):

PR #1060, opened 2024-09-18
[API] Add endpoints to list KafkaRebalances and patch action annotation

MikeEdgar:


MikeEdgar:
@hemahg , I'm not sure having `action` in `attributes` for handling updates is a great approach. This may be better represented by PATCHing this information in another `meta` entry. Either way, it's fairly minor and the majority of how the endpoints look and feel will remain the same.

hemahg (approved):

PR #1070, opened 2024-09-24
feat: kafka Rebalance list

hemahg:
feat: kafka Rebalance list

Kafka rebalance list 
Screenshot 2024-10-01 at 11 37 42 AM

optimization Result modal
Screenshot 2024-10-01 at 11 38 14 AM


MikeEdgar (commented):
Can the headers/values for the status counts be centered? The ones for the total and stopped seem to be in a weird position.
image

Similarly for the mode/auto-approval/brokers data:
image

I think if there are no brokers returned by the API (which should be the case when the mode=full), we can display "N/A" or something.

When entries in the proposal report are missing values like the "excluded" labels here, should we put "None" or "-" as a placeholder? 
image

Actions when state is "ProposalReady" should be "Approve" and "Refresh". I'll look through the code and see if I can spot why it's not correct.
image

ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/rebalances/RebalanceTable.tsx
```diff
@@ -0,0 +1,294 @@
...
+            );
+        }
+      }}
+      renderActions={({ row, ActionsColumn }) => (
+        <ActionsColumn
+          items={[
+            {
+              title: t("approve"),
+              onClick: () => onApprove(row),
+              isDisabled: row.meta?.allowedActions.includes("approve"),
```
Comment (MikeEdgar) [inline ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/rebalances/RebalanceTable.tsx:198]:
```suggestion
              isDisabled: !row.meta?.allowedActions.includes("approve"),
```

ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/rebalances/RebalanceTable.tsx
```diff
@@ -0,0 +1,294 @@
...
+            },
+            {
+              title: t("refresh"),
+              onClick: () => onRefresh(row),
```
Comment (MikeEdgar) [inline ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/rebalances/RebalanceTable.tsx:202]:
```suggestion
              onClick: () => onRefresh(row),
              isDisabled: !row.meta?.allowedActions.includes("refresh"),
```

ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/rebalances/RebalanceTable.tsx
```diff
@@ -0,0 +1,294 @@
...
+            },
+            {
+              title: t("stop"),
+              onClick: () => onStop(row),
+              isDisabled: !row.meta?.allowedActions.includes("approve"),
```
Comment (MikeEdgar) [inline ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/rebalances/RebalanceTable.tsx:207]:
```suggestion
              isDisabled: !row.meta?.allowedActions.includes("stop"),
```

MikeEdgar (changes requested):
A few suggestions for the options menu

MikeEdgar (changes requested):
When filtering the list and nothing matches the filter, the empty state is shown without a way to clear the filter. Can this view incorporate the search box also?

image

ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/rebalances/RebalanceTable.tsx
```diff
@@ -0,0 +1,294 @@
...
+          onRemoveChip: (status) => {
+            const newStatus = (filterStatus || []).filter((s) => s !== status);
+            onFilterStatusChange(newStatus);
+          },
+          onRemoveGroup: () => {
+            onFilterStatusChange(undefined);
+          },
+          options: StatusLabel,
+        },
+        mode: {
```
Comment (MikeEdgar) [inline ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/rebalances/RebalanceTable.tsx:278]:
```suggestion
        Mode: {
```

Also, can mode function like the status search with a fixed list of values that can be chosen with select boxes? There are only three values: `full`, `add-brokers`, and `remove-brokers`.

MikeEdgar (commented):
Refreshing the page on the rebalance tab, the "Brokers" header and "Overview"/"Rebalance" tabs are lost:
image

ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/rebalances/RebalanceTable.tsx
Comment (hemahg) [reply ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/rebalances/RebalanceTable.tsx:207]: I have implemented the logic according to the design: if 'approve' is present in the allowedActions list, the 'Stop' button will be enabled. If 'approve' is absent in the allowed actions, the 'Stop' button will be disabled, and the 'Approve' button will be enabled.

Comment (hemahg) [reply ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/rebalances/RebalanceTable.tsx:207]: Screenshot 2024-10-03 at 4 37 47 PM

ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/rebalances/RebalanceTable.tsx
Comment (hemahg) [reply ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/rebalances/RebalanceTable.tsx:306]: updated

ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/rebalances/RebalanceTable.tsx
Comment (MikeEdgar) [reply ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/rebalances/RebalanceTable.tsx:223]: > if 'approve' is present in the allowedActions list, the 'Stop' button will be enabled. If 'approve' is absent in the allowed actions, the 'Stop' button will be disabled, and the 'Approve' button will be enabled.

I don't think these make sense. The Stop option should only be enabled when `stop` is one of the `allowedActions`. Similarly, the Approve option should be enabled when `approve` is one of the `allowedActions`. Also, the same for the Refresh option. These are driven by the current state, as defined by Strimzi's [`KafkaRebalanceState`](https://github.com/strimzi/strimzi-kafka-operator/blob/637d93483ba1ff1a9277e437f04b40d31e09b0ba/api/src/main/java/io/strimzi/api/kafka/model/rebalance/KafkaRebalanceState.java#L9)

Comment (hemahg) [reply ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/rebalances/RebalanceTable.tsx]: updated

MikeEdgar (approved):
Looks good, thanks @hemahg

PR #1103, opened 2024-10-07
[API] Add endpoint to patch Kafka resource pause annotation

MikeEdgar:
Add endpoint to patch Kafka resource pause annotation. When the paused annotation is present on the Kafka CR, the kafka resource returned by the API will include a `.meta.reconciliationPaused` boolean flag that is `true` or `false`, depending on the annotation value. Otherwise, if there is no annotation on the Cr, the `.meta.reconciliationPaused` will not be present.


MikeEdgar:
@hemahg - these are the API changes to pause/resume the Kafka resource reconciliation.

hemahg (approved):
yes @MikeEdgar, everything was working as expected

PR #1109, opened 2024-10-09
Apicurio Registry integration with Avro/Protobuf ser/des support

MikeEdgar:
This PR adds a multiformat message deserializer and serializer (not exposed in UI) and support to integrate with Apicurio Registry for retrieval of message schemas. The deserializer works generally by locating a schema identifier in a message's headers or directly in the key/value being deserialized. The identifier is used to fetch a schema from Apicurio Registry and either Avro or Protobuf deserialization is performed based on the type of schema returned.

Failure to locate a schema for a discovered identifier is communicated to the client (UI). When successful, the schema's name and link (to retrieve the content) will be returned to the client so that a user may view the schema if wanted. Retrieval of the schema content is proxied to Apicurio Registry via a new `SchemasResource` added in this PR as well.

Inline information for key/value formats:
image

Inline schema resolution warnings:
image

### DIFF
--- a/api/src/main/java/com/github/streamshub/console/api/ClientFactory.java
+++ b/api/src/main/java/com/github/streamshub/console/api/ClientFactory.java
@@ -1,9 +1,5 @@
 package com.github.streamshub.console.api;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.UncheckedIOException;
-import java.nio.file.Path;
 import java.util.Base64;
 import java.util.Collection;
 import java.util.Collections;
@@ -55,21 +51,17 @@ import org.apache.kafka.common.security.scram.ScramLoginModule;
 import org.apache.kafka.common.serialization.ByteArrayDeserializer;
 import org.apache.kafka.common.serialization.StringSerializer;
 import org.eclipse.microprofile.config.Config;
-import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.jboss.logging.Logger;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import com.github.streamshub.console.api.support.Holder;
 import com.github.streamshub.console.api.support.KafkaContext;
 import com.github.streamshub.console.api.support.TrustAllCertificateManager;
-import com.github.streamshub.console.api.support.ValidationProxy;
 import com.github.streamshub.console.api.support.serdes.RecordData;
 import com.github.streamshub.console.config.ConsoleConfig;
 import com.github.streamshub.console.config.KafkaClusterConfig;
+import com.github.streamshub.console.config.SchemaRegistryConfig;
 
-import io.apicurio.registry.rest.client.RegistryClient;
-import io.apicurio.registry.rest.client.RegistryClientFactory;
 import io.apicurio.registry.serde.SerdeConfig;
 import io.fabric8.kubernetes.client.informers.ResourceEventHandler;
 import io.fabric8.kubernetes.client.informers.SharedIndexInformer;
@@ -128,14 +120,10 @@ public class ClientFactory {
     ObjectMapper mapper;
 
     @Inject
-    @ConfigProperty(name = "console.config-path")
-    Optional<String> configPath;
+    ConsoleConfig consoleConfig;
 
     @Inject
     Holder<SharedIndexInformer<Kafka>> kafkaInformer;
-
-    @Inject
-    ValidationProxy validationService;
 
     @Inject
     Instance<TrustAllCertificateManager> trustManager;
@@ -168,57 +156,12 @@ public class ClientFactory {
 
     @Produces
     @ApplicationScoped
-    public ConsoleConfig produceConsoleConfig() {
-        return configPath.map(Path::of)
-            .map(Path::toUri)
-            .map(uri -> {
-                try {
-                    return uri.toURL();
-                } catch (IOException e) {
-                    throw new UncheckedIOException(e);
-                }
-            })
-            .filter(Objects::nonNull)
-            .map(url -> {
-                log.infof("Loading console configuration from %s", url);
-                ObjectMapper yamlMapper = mapper.copyWith(new YAMLFactory());
-
-                try (InputStream stream = url.openStream()) {
-                    return yamlMapper.readValue(stream, ConsoleConfig.class);
-                } catch (IOException e) {
-                    throw new UncheckedIOException(e);
-                }
-            })
-            .map(consoleConfig -> {
-                consoleConfig.getKafka().getClusters().stream().forEach(cluster -> {
-                    if (cluster.getSchemaRegistry() != null) {
-                        var registryConfig = cluster.getSchemaRegistry();
-                        registryConfig.setUrl(resolveValue(registryConfig.getUrl()));
-                    }
-                    resolveValues(cluster.getProperties());
-                    resolveValues(cluster.getAdminProperties());
-                    resolveValues(cluster.getProducerProperties());
-                    resolveValues(cluster.getConsumerProperties());
-                });
-
-                return consoleConfig;
-            })
-            .map(validationService::validate)
-            .orElseGet(() -> {
-                log.warn("Console configuration has not been specified using `console.config-path` property");
-                return new ConsoleConfig();
-            });
-    }
-
-    @Produces
-    @ApplicationScoped
-    Map<String, KafkaContext> produceKafkaContexts(ConsoleConfig consoleConfig,
-            Function<Map<String, Object>, Admin> adminBuilder) {
+    Map<String, KafkaContext> produceKafkaContexts(Function<Map<String, Object>, Admin> adminBuilder) {
 
         final Map<String, KafkaContext> contexts = new ConcurrentHashMap<>();
 
         if (kafkaInformer.isPresent()) {
-            addKafkaEventHandler(contexts, consoleConfig, adminBuilder);
+            addKafkaEventHandler(contexts, adminBuilder);
         }
 
         // Configure clusters that will not be configured by events
@@ -236,7 +179,6 @@ public class ClientFactory {
     }
 
     void addKafkaEventHandler(Map<String, KafkaContext> contexts,
-            ConsoleConfig consoleConfig,
             Function<Map<String, Object>, Admin> adminBuilder) {
 
         kafkaInformer.get().addEventHandlerWithResyncPeriod(new ResourceEventHandler<Kafka>() {
@@ -348,15 +290,18 @@ public class ClientFactory {
                 admin = adminBuilder.apply(adminConfigs);
             }
 
-            RegistryClient schemaRegistryClient = null;
+            SchemaRegistryConfig registryConfig = null;
 
             if (clusterConfig.getSchemaRegistry() != null) {
-                String registryUrl = clusterConfig.getSchemaRegistry().getUrl();
-                schemaRegistryClient = RegistryClientFactory.create(registryUrl); // NOSONAR - closed elsewhere
+                registryConfig = consoleConfig.getSchemaRegistries()
+                        .stream()
+                        .filter(registry -> registry.getName().equals(clusterConfig.getSchemaRegistry()))
+                        .findFirst()
+                        .orElseThrow();
             }
 
             KafkaContext ctx = new KafkaContext(clusterConfig, kafkaResource.orElse(null), clientConfigs, admin);
-            ctx.schemaRegistryClient(schemaRegistryClient, mapper);
+            ctx.schemaRegistryClient(registryConfig, mapper);
 
             KafkaContext previous = contexts.put(clusterId, ctx);
 
@@ -560,7 +505,7 @@ public class ClientFactory {
 
     @Produces
     @RequestScoped
-    public Consumer<RecordData, RecordData> consumerSupplier(ConsoleConfig consoleConfig, KafkaContext context) {
+    public Consumer<RecordData, RecordData> consumerSupplier(KafkaContext context) {
         var configs = maybeAuthenticate(context, Consumer.class);
 
         return new KafkaConsumer<>(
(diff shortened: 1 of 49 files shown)


MikeEdgar:
@k-wall FYI, you may be interested in case it is relevant to schema usage for Kroxylicious.

k-wall:
Thanks @MikeEdgar, I'll review. @SamBarker FYI

MikeEdgar:
@carlesarnal if you have some time to look through these changes and let me know if anything looks like it could be improved, that would be great :smile: This is the PR to incorporate Apicurio Registry to generically deserialize Avro or Protobuf in our Kafka console, as well as give the UI a handle to fetch the schema for display.

api/src/main/java/com/github/streamshub/console/api/ClientFactory.java
```diff
@@ -438,6 +451,7 @@ Map<String, Object> requiredConsumerConfig() {
         configs.put(ConsumerConfig.SESSION_TIMEOUT_MS_CONFIG, 50_000);
         configs.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
         configs.put(ConsumerConfig.DEFAULT_API_TIMEOUT_MS_CONFIG, 5000);
+        configs.put(SerdeConfig.ENABLE_HEADERS, "true");
```
k-wall, line 454: https://github.com/Apicurio/apicurio-registry/pull/5149 disables headers by default. Is Console aware of that? Is Console in sympathy with that?

api/src/main/java/com/github/streamshub/console/api/ClientFactory.java
```diff
@@ -552,26 +566,31 @@ public void disposeKafkaContext(@Disposes KafkaContext context, Map<String, Kafk
...
+
+        return new KafkaConsumer<>(
+                configs,
+                context.schemaRegistryContext().keyDeserializer(),
+                context.schemaRegistryContext().valueDeserializer());
     }
 
-    public void consumerDisposer(@Disposes Supplier<Consumer<byte[], byte[]>> consumer) {
-        consumer.get().close();
+    public void disposeConsumerSupplier(@Disposes Consumer<RecordData, RecordData> consumer) {
```
k-wall, line 578: Why the word Supplier in this method name?

api/src/main/java/com/github/streamshub/console/api/SchemasResource.java
```diff
@@ -0,0 +1,93 @@
...
+            @PathParam("schemaId")
+            String schemaId) {
+
+        int clusterTerm = schemaId.indexOf('.');
+
+        if (clusterTerm < 0) {
+            throw new NotFoundException("No such schema");
+        }
+
+        String clusterId = new String(Base64.getUrlDecoder().decode(schemaId.substring(0, clusterTerm)));
```
k-wall, line 62: I'm used to see REST libraries perform the decode for you.  Are we double encoding it or something?

api/src/main/java/com/github/streamshub/console/api/ClientFactory.java
MikeEdgar: The code will currently check the headers first, then fall back to the embedded ID in the message for de-serialization. 

This configuration is more significant for the message producer, which isn't yet part of the UI, so the risk is probably low. I would say we can remove this option here and defer how to handle headers vs. embedded schema identifiers until we add the ability to produce messages to the UI. There may end up being options in the UI to choose where to put the ID, which schema to use, etc.

api/src/main/java/com/github/streamshub/console/api/ClientFactory.java
MikeEdgar: I'll remove it. The name is a relic of the old type being `Supplier<Consumer<x, x>>`

api/src/main/java/com/github/streamshub/console/api/SchemasResource.java
```diff
@@ -0,0 +1,93 @@
...
+
+        if (!kafkaContexts.containsKey(clusterId)) {
+            throw new NotFoundException("No such schema");
```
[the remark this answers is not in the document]
MikeEdgar: This endpoint is meant to be non-hierarchical, with the Kafka cluster Id being part of the opaque schema Id. I'm thinking now that's confusing and doesn't really buy us anything, so I'll probably switch it to `/api/kafkas/{kafkaId}/schemas/{schemaId}`. Basically, my intent was to disassociate a Kafka instance from a schema since they aren't related in Apicurio. However, in the console we need the Kafka Id to find the Apicurio API URL/client.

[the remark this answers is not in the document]
MikeEdgar: To answer your question, even though internally it can't find the Kafka cluster, from the perspective of a client to `api/schemas/{schemaId}`, a 404 means the schema wasn't found, regardless of the implementation of how to find it.

api/src/main/java/com/github/streamshub/console/api/SchemasResource.java
```diff
@@ -0,0 +1,93 @@
...
+    @Produces(MediaType.APPLICATION_JSON)
+    @APIResponse(responseCode = "200", ref = "Configurations", content = @Content())
+    @APIResponse(responseCode = "404", ref = "NotFound")
+    @APIResponse(responseCode = "500", ref = "ServerError")
+    @APIResponse(responseCode = "504", ref = "ServerTimeout")
+    public Response getSchemaContent(
+            @Parameter(description = "Schema identifier")
+            @PathParam("schemaId")
+            String schemaId) {
+
```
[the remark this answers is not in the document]
MikeEdgar: See my other comment about making it hierarchical and splitting the Kafka ID out from schema Id. I'll also add a note about the schemaId being a serialized Apicurio `ArtifactReference`.

api/src/main/java/com/github/streamshub/console/api/ClientFactory.java
MikeEdgar: After further review, I think we should keep this option enabled. Without it, deserialization will not check the headers at all, which limits the console's ability to decode records written by an application that does use the headers.

api/src/main/java/com/github/streamshub/console/api/SchemasResource.java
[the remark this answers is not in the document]
k-wall: I'm wondering whether the model of associating a schema registry to a kafka cluster is the right one.  In Kafka, other the client knows about the schema registry.  It is possible for topic A to use a schema hosted on one registry and topic B to use a schema on different.

I don't have a concrete suggestion to make right now but wanted to express unease. Did we research how other products/vendors approach this problem?  Are we going in the right direction on this?

I suppose we are talking about eventually having some kind of topic metadata service.  Once such a thing exists, it could identify the topic's schema and the registry that holds that schema.

api/src/main/java/com/github/streamshub/console/api/model/HasLinks.java
```diff
@@ -0,0 +1,26 @@
...
+        @SuppressWarnings("unchecked")
+        T t = (T) this;
+        return t;
+    }
+
+    private static <K, V> Map<K, V> addEntry(Map<K, V> map, K key, V value) {
+        if (map == null) {
+            map = new LinkedHashMap<>();
+        }
+        map.put(key, value);
```
k-wall, line 23: do we need to think about the case where an implementation is using a immutable map?

api/src/main/java/com/github/streamshub/console/api/support/serdes/RecordData.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public com.github.streamshub.console.api.model.Error error() {
+        return error;
+    }
+
+    public String dataString(Integer maxValueLength) {
+        return bytesToString(data, maxValueLength);
+    }
+
+    public static String bytesToString(byte[] bytes, Integer maxValueLength) {
```
k-wall, line 45: why use an object rather than primitive?

api/src/main/java/com/github/streamshub/console/api/support/KafkaContext.java
```diff
@@ -115,4 +158,78 @@ public Optional<String> tokenUrl() {
...
+        private final RegistryClient registryClient;
+        private final MultiformatDeserializer keyDeserializer;
+        private final MultiformatDeserializer valueDeserializer;
+        private final MultiformatSerializer keySerializer;
+        private final MultiformatSerializer valueSerializer;
+
+        SchemaRegistryContext(RegistryClient schemaRegistryClient) {
+            this.registryClient = schemaRegistryClient;
+
+            ObjectMapper objectMapper = new ObjectMapper();
```
k-wall, line 179: Could this be a constant?

api/src/main/java/com/github/streamshub/console/api/support/serdes/MultiformatSchemaParser.java
```diff
@@ -0,0 +1,122 @@
...
+ * This class can likely be improved with more a advanced/accurate detection
+ * process.
+ */
+public class MultiformatSchemaParser<D> implements SchemaParser<Object, D> {
+
+    private static final int CACHE_LIMIT = 20;
+    private static final char[] PROTOBUF_INTRO = "message ".toCharArray();
+
+    private final Map<String, SchemaParser<?, ?>> delegates;
+    private final Map<byte[], Object> schemaCache = new LinkedHashMap<>(CACHE_LIMIT);
```
k-wall, line 32: Thread safety?  LHM aren't thread safe IIRC

k-wall: I wonder if you'd be better with something like Caffeine.

api/src/main/java/com/github/streamshub/console/api/support/serdes/MultiformatSchemaParser.java
```diff
@@ -0,0 +1,122 @@
...
+import io.apicurio.registry.resolver.SchemaParser;
+import io.apicurio.registry.resolver.data.Record;
+import io.apicurio.registry.types.ArtifactType;
+
+/**
+ * Schema parser that delegates to either the Avro or Protobuf schema parser.
+ * This class attempts to sniff the schema's type using a primitive check for a
+ * Protobuf message header.
+ *
+ * This class can likely be improved with more a advanced/accurate detection
```
k-wall, line 23: typo

api/src/main/java/com/github/streamshub/console/api/support/serdes/MultiformatSchemaParser.java
```diff
@@ -0,0 +1,122 @@
...
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+    }
+
+    @Override
+    public String artifactType() {
+        throw new UnsupportedOperationException("MultiformatSchemaParser#artifactType()");
+    }
+
+    @Override
+    public Object parseSchema(byte[] rawSchema, Map<String, ParsedSchema<Object>> resolvedReferences) {
```
k-wall, line 46: worthwhile javadoc'ing that this may return null

k-wall:
@MikeEdgar I've given the code a rather superficial review, no major red flags stand out.  Good work.  I'm not going to be able to go deeper before I leave for PTO.

api/src/main/java/com/github/streamshub/console/api/SchemasResource.java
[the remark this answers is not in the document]
MikeEdgar: > Did we research how other products/vendors approach this problem?

Both AKHQ [1] and kafka-ui [2] do it per cluster. I agree that having the metadata storage for topics would make a difference in how we approach this. Associating them to Kafka clusters also doesn't feel ideal to me, but I wonder how often multiple registries are being used in an environment. The only way I can think to do that (without metadata storage) would be to allow for per-topic configuration of the registry, which could possibly be a future change if someone asks for it.

[1] https://akhq.io/docs/configuration/brokers.html
[2] https://docs.kafka-ui.provectus.io/configuration/serialization-serde#schemaregistry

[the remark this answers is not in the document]
k-wall: > Associating them to Kafka clusters also doesn't feel ideal to me, but I wonder how often multiple registries are being used in an environment. 

I don't know that either.

The approach you are taking seems like a reasonable place to start.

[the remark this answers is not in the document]
MikeEdgar: @k-wall another option that might work is to allow for schema registries to be configured at the top level and then individual Kafka clusters could reference them. E.g.

```yaml
schemaRegistries:
  - name: production-registry
    url: http://prod.example.com/...
  - name: staging-registry
    url: http://staging.example.com...

kafka:
  clusters:
    - name: prod-k1
      schemaRegistry: production-registry
    - name: prod-k2
      schemaRegistry: production-registry
    - name: staging-k3
      schemaRegistry: staging-registry
```

This would make the `SchemasResource` URL a more related to reality: `/api/schema-registries/{registryName}/schemas/{schemaId}` and make the configuration a tiny bit DRYer. We're still binding a single registry per Kafka cluster, but the config for each registry is "reusable" within the config.

wdyt?

api/src/main/java/com/github/streamshub/console/api/model/HasLinks.java
MikeEdgar: Not in this case. We're in control of the implementations and there shouldn't be a need for them to be immutable. These are only used to build REST responses.

api/src/main/java/com/github/streamshub/console/api/support/KafkaContext.java
MikeEdgar: Replaced this ObjectMapper with a provided application-scoped mapper.

api/src/main/java/com/github/streamshub/console/api/support/serdes/RecordData.java
MikeEdgar: Added Javadocs for the RecordData string methods, including `maxValueLength` argument.

api/src/main/java/com/github/streamshub/console/api/SchemasResource.java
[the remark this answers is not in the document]
MikeEdgar: I've gone ahead with this change. As mentioned, this will reduce duplicate configurations between Kafka clusters and also establishes a pattern that should be extended to Prometheus instance configurations and how they are bound to a Kafka cluster for metrics retrieval.

hemahg (approved):

PR #1111, opened 2024-10-10
feat: pause reconciliation

hemahg:
feat: pause reconciliation

Design [link](https://www.figma.com/proto/JHubc4ULXXhKAEk8VyaW2f/AMQ-Streams-console?page-id=398%3A9817&node-id=1100-7414&node-type=FRAME&viewport=303%2C-232%2C0.29&t=ufkbHniczdcX05ET-1&scaling=min-zoom&content-scaling=fixed)

### DIFF
--- a/ui/components/AppLayout.tsx
+++ b/ui/components/AppLayout.tsx
@@ -1,12 +1,4 @@
-import {
-  Banner,
-  Bullseye,
-  Button,
-  Flex,
-  FlexItem,
-  Nav,
-  Page,
-} from "@/libs/patternfly/react-core";
+import { Nav, Page } from "@/libs/patternfly/react-core";
 import { useTranslations } from "next-intl";
 import { PropsWithChildren, ReactNode } from "react";
 import { AppMasthead } from "./AppMasthead";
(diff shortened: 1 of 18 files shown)


MikeEdgar (commented):
At root page of the application I am seeing repeated HTTP 400 errors. It looks like the UI is calling getKafkaCluster to fetch a single cluster with a blank/empty `kafkaId`.

ui/components/ClusterOverview/ClusterCard.tsx
```diff
@@ -218,6 +247,69 @@ export function ClusterCard({
                         </DataListItemRow>
                       </DataListItem>
                     )}
+                    {isReconciliationPaused && (
+                      <DataListItem aria-labelledby={`reconciliation-warning`}>
```
MikeEdgar, line 251: The `messages` array already contains a condition for `ReconciliationPaused`. Can we use that instead of adding an additional row? It results in the generated row plus the one that Strimzi gives us for `ReconciliationPaused`.

image

hemahg: fixed

ui/components/AppLayout.tsx
```diff
@@ -28,7 +37,14 @@ export function AppLayout({
...
       <ClusterDrawerProvider>
-        <ClusterDrawer>{children}</ClusterDrawer>
+        {isValidKafkaId ? (
+          <ReconciliationProvider kafkaId={kafkaId}>
+            <ReconciliationPausedBanner kafkaId={kafkaId} />
+            <ClusterDrawer>{children}</ClusterDrawer>
+          </ReconciliationProvider>
+        ) : (
+          <></>
+        )}
```
MikeEdgar, line 47: This change allows the list of multiple Kafkas to still appear, but blocks the banner if there is no `kafkaId`.

```suggestion
        <ReconciliationProvider kafkaId={kafkaId ?? ""}>
          {isValidKafkaId && (
            <ReconciliationPausedBanner kafkaId={kafkaId} />
          )}
          <ClusterDrawer>{children}</ClusterDrawer>
        </ReconciliationProvider>
```

MikeEdgar (approved):
One more fix, then good with me. Thanks @hemahg 

There is another improvement that can be made in a follow-up PR to hide the pause button when a Kafka cluster's `.meta.managed` field is `false`. These Kafka clusters do not have a (known) Kafka CR so we can't pause/resume them.

PR #1121, opened 2024-10-18
Obtain topic/partition overview counts from topics API, not metrics

MikeEdgar:
Remove the retrieval of summary topic/partition counts from Prometheus and replace with using the summary information available in the topic list response. Using metrics is buggy and the "offline" count will never be correct anyway due to not enough information being available in Prometheus.

Before:
image

After:
image


MikeEdgar:
@jankalinic FYI

ui/app/[locale]/(authorized)/kafka/[kafkaId]/overview/page.tsx
```diff
@@ -25,6 +25,7 @@ export default function OverviewPage({ params }: { params: KafkaParams }) {
     "outgoingByteRate",
     "incomingByteRate",
   ]);
+  const topics = getTopics(params.kafkaId, { fields: "status", pageSize: 1 });
```
Comment (MikeEdgar) [inline ui/app/[locale]/(authorized)/kafka/[kafkaId]/overview/page.tsx:28]: Only select the minimum number of records because we only care about the summary meta data for this request, not the list of topics.

jankalinic (approved):
Thank you!

katheris (approved):
LGTM

PR #1126, opened 2024-10-21
Authorization, audit logging, optional OIDC authN, UI error handling

MikeEdgar:
This change includes many updates to both the API and UI layers in order to support configuration of authorization rules, optionally integrate with an OIDC provider, improve error handling in the UI, and perform audit logging for decisions based on the security rules.

- Security may be configured via the Console configuration YAML (example in examples/console-config.yaml in this PR). The configuration has two "levels" - global and per-Kafka. The global level is the `security` property at the root of the YAML where OIDC is optionally configured, OIDC subjects are defined, and Kafka cluster-level access is granted. Security at the Kafka level (within the `kafka.clusters` array entries) is where per-cluster access is granted to the topics, consumer groups, etc. for each cluster.
- Auditing is enabled based on decision type and may include only specific resources or resources matching a name glob. For example, this configuration will audit at INFO level any time access to a consumer group is denied, regardless of the operation (given by privileges).
  ```yaml
  security:
    audit:
      - decision: denied
        resources:
        - consumerGroups
        privileges: [ "*" ]
  ```
- UI errors will now be handled within the same component where they occur, removing the old redirect behavior that existed in many places previously. When a user lacks authorization to the data in a particular view, that information will be displayed accordingly.

What's missing?
- Support to configure security via the operator, #1247 
- Support for self-signed certificates used by the OIDC provider (without generally disabling TLS verification). This will be done via #1248


MikeEdgar:
@katheris and I discussed this PR and agreed that we can go ahead with merging it. As noted in the description there are (at least) two follow-up features with #1247 and #1248 that need this.

PR #1153, opened 2024-10-28
feat: hide pause button when Kafka cluster is unmanaged

hemahg:
feat: hide pause button when Kafka cluster is unmanaged


MikeEdgar (approved):
Looks good to me, thanks @hemahg

PR #1191, opened 2024-11-07
Upgrade to Patternfly v6

hemahg:
Upgrade to Patternfly v6


MikeEdgar (commented):
Looks nice @hemahg. Here are a few issues I've seen browsing around the app so far. Also, if it is simple, it might be a good option to allow a toggle for light/dark modes [1] like 
image


[1] https://www.patternfly.org/developer-resources/dark-theme-handbook

---

Brokers heading shifted up and left. Also, the rebalance tab is missing and this image is for a managed Kafka with a CR.
image

Compared to Topics:
image

---

Same heading issue for an individual topic, and the selected tab isn't highlighted:
image

---

Partition tabs are transparent - scrolling the configuration is visible through the tabs:
image

hemahg:
1.Fixing the Navigation and Highlight Issues: The heading issue in the Topic and Broker tabs has been addressed.
Screenshot 2024-11-18 at 12 47 01 PM
2.Regarding the rebalance tab, I couldn't reproduce the missing tab issue.
Screenshot 2024-11-18 at 12 47 01 PM
3.In the latest PatternFly version, the navigation highlights in white if the page section background is gray [PatternFly Horizontal Nav Demo](https://www.patternfly.org/components/navigation/react-demos/horizontal-nav/). However, since the page section is white, tab changes aren't visually highlighted. The previously used component is now deprecated:

MikeEdgar:
Ignore what I said earlier about the missing rebalance tab - it was an issue with my configuration. Sorry for any confusion on that.

riccardo-forina:
At the time, the tabs were full client components that did not support the routing at all. We have a unique url tied to each tab that loads the new portion of data following the way next.js router works. So I couldn't use that. And the tabs weren't a great fit estically speaking.

Feel free to change them to whatever works better.

MikeEdgar:
Ok, I see what you mean. It looks like there is support in PF6 for "tabs linked to Nav elements". We can think about whether that makes sense for a future PR.

https://www.patternfly.org/components/tabs#tabs-linked-to-nav-elements

riccardo-forina:
That option was already available, but it renders plain html links, not Next Link components, unfortunately. 

I suppose we could still use that, and add an onClick handler that triggers the react navigation programmatically preventing the default behavior. Not ideal, but workable I suppose

PR #1192, opened 2024-11-07
Set Kafka `group.initial.rebalance.delay.ms` to `0` in tests

### LINKED ISSUE #1175: Investigate consumer group IT performance
The run time for the consumer group integration test seems to have increased after #1149 . Investigate and see if it can be improved.

MikeEdgar:
Fixes #1175


MikeEdgar:
@katheris can you do a quick review for this change? As noted in an earlier comment, it's to account for a change in the Strimzi test-container that impacts the test runtime for the console.

katheris (approved):

PR #1212, opened 2024-11-14
fix: Improve slow consume records operation

MikeEdgar:
- Skip consumer assignments for topic partitions with no data
- Skip consumer assignments that seek beyond the end of the assigned topic partition
- Limit consumer poll time to 100ms per request to poll the consumer
- Make total time polling via the endpoint configurable, default 5s
- Fix NPE in multi-format de-serializer


MikeEdgar:
@hemahg this may help with the timeouts you receive using the `topics/{topicId}/records` endpoint.

hemahg (approved):
Thank you @MikeEdgar

PR #1226, opened 2024-11-20
fix: common module validation and rm duplicate cluster check

### LINKED ISSUE #1225: Validations not running properly for classes in `common` module
The validation constraints are not being applied for classes in the `common` module. This may be fixed by making sure the module contains a Jandex index so that the annotations are discovered by the runtime.

MikeEdgar:
Fixes #1225 

Duplicate cluster check is verified via constraint annotations and is no longer needed in the `ClientFactory`.


MikeEdgar:
@katheris please review if you have a chance. This was in the auth PR, but I'm trying to minimize it since it's already enormous. This removes a duplicate cluster check when creating the Kafka clients because it would have been already validated when the configurations are loaded at startup. The addition of the jandex index makes the validation work properly.

PR #1227, opened 2024-11-20
Set limited securityContext for all containers

MikeEdgar:


jankalinic (approved):
Looks good, let me run some tests for confirmation.

jankalinic:
Tested against my rapidast tests. Works. Thank you

PR #1230, opened 2024-11-21
[OLM] Operator name change

jankalinic:
This PR targets current generic `console-operator` naming and changes it to `streamshub-operator`. This change affects kubernetes resources like Operator, CSV, Pod, Deployment, but also a ServiceAccount. However this change ignores image names as their full name already contains `streamshub` part in the registry name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ spec:
 ```
 
 ### Deploy the operator directly
-Deploying the operator without the use of OLM requires applying the component Kubernetes resources for the operator directly. These resources are bundled and attached to each StreamsHub Console release. The latest release can be found [here](https://github.com/streamshub/console/releases/latest). The resource file is named `console-operator-x.y.z.yaml` where `x.y.z` is the released version.
+Deploying the operator without the use of OLM requires applying the component Kubernetes resources for the operator directly. These resources are bundled and attached to each StreamsHub Console release. The latest release can be found [here](https://github.com/streamshub/console/releases/latest). The resource file is named `streamshub-console-operator-x.y.z.yaml` where `x.y.z` is the released version.
 
 This example will create the operator's resources in the `default` namespace. Modify the `NAMESPACE` variable according to your needs and set `VERSION` to the [latest release](https://github.com/streamshub/console/releases/latest).
 ```
--- a/README.md
+++ b/README.md
@@ -96,13 +96,13 @@ spec:
 ```
 
 ### Deploy the operator directly
-Deploying the operator without the use of OLM requires applying the component Kubernetes resources for the operator directly. These resources are bundled and attached to each StreamsHub Console release. The latest release can be found [here](https://github.com/streamshub/console/releases/latest). The resource file is named `streamshub-console-operator-x.y.z.yaml` where `x.y.z` is the released version.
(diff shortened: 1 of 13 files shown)


operator/src/main/java/com/github/streamshub/console/LeaderConfiguration.java
```diff
@@ -7,6 +7,6 @@
 @ApplicationScoped
 public class LeaderConfiguration extends LeaderElectionConfiguration {
     public LeaderConfiguration() {
-        super("console-operator-lease");
+        super("streamshub-console-operator-lease");
```
jankalinic, line 10: Not sure if this is needed or not, but i'll leave it up to review.

operator/src/main/resources/application.properties
```diff
@@ -6,7 +6,7 @@ console.deployment.default-ui-image=quay.io/streamshub/console-ui:${console.depl
 
 quarkus.container-image.build=true
 #quarkus.container-image.group=
-quarkus.container-image.name=console-operator
+quarkus.container-image.name=streamshub-console-operator
```
MikeEdgar, line 9: This line changes the name of the image that is pushed. That is, `quay.io/streamshub/console-operator` becomes `quay.io/streamshub/streamshub-console-operator`.

Is it necessary since the scripting changes will modify the name in the YAMLs?

operator/src/main/java/com/github/streamshub/console/LeaderConfiguration.java
MikeEdgar: I think this should be fine

operator/src/main/resources/application.properties
jankalinic: Good point, it should remain console-operator as you said.

README.md
```diff
@@ -96,7 +96,7 @@ spec:
 ```
 
 ### Deploy the operator directly
-Deploying the operator without the use of OLM requires applying the component Kubernetes resources for the operator directly. These resources are bundled and attached to each StreamsHub Console release. The latest release can be found [here](https://github.com/streamshub/console/releases/latest). The resource file is named `console-operator-x.y.z.yaml` where `x.y.z` is the released version.
+Deploying the operator without the use of OLM requires applying the component Kubernetes resources for the operator directly. These resources are bundled and attached to each StreamsHub Console release. The latest release can be found [here](https://github.com/streamshub/console/releases/latest). The resource file is named `streamshub-console-operator-x.y.z.yaml` where `x.y.z` is the released version.
```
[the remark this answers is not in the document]
jankalinic: I guess it will not matter if we have x.y.z in the file name since its already under the release folder / tag.
It might add some clarity once the file is lets say somewhere on someones local drive and it's just called console-operator.yaml. 
So i am neutral with this one, maybe leaning a bit into removing x.y.z.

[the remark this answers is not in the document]
jankalinic: Nah, lets just remove it 😆 . It looks better that way.
`https://github.com/streamshub/console/releases/download/0.3.3/streamshub-console-operator.yaml`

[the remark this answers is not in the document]
MikeEdgar: image

MikeEdgar (approved):
Looks good, thanks @jankalinic

PR #1245, opened 2024-11-26
Provide a way to override operand image, env, and resources in CRD

MikeEdgar:
This change deprecates the existing `images` and `env` properties of the CRD and replaces them with per-container objects to specify a container image, env, and resources (cpu/memory).

Currently in the Console CR, API container ENVs and override images can be given like
```yaml
spec:
  env:
  - name: SOME_ENVVAR
    value: "custom-value"
  images:
    api: example.com/myorg/my-api-image:1.0
    ui: example.com/myorg/my-ui-image:1.0
```

The new way with the change is:
```yaml
spec:
  containers:
    api:
      spec:
        image: example.com/myorg/my-api-image:1.0
        resources:
          limits:
            cpu: 500m
            memory: 256Mi
        env:
        - name: SOME_ENVVAR
          value: "custom-value"
    ui:
      spec:
        image: example.com/myorg/my-ui-image:1.0
        resources:
          limits:
            cpu: 500m
            memory: 256Mi
        env:
        - name: SOME_ENVVAR_UI
          value: "custom-value"
```

Both sets of properties are still mapped to the console deployment, but the old properties are marked as deprecated in the CRD descriptions.


MikeEdgar:
@k-wall, I'm curious to get your thoughts/feedback on the proposed CRD format in the description.

jankalinic (approved):
Thanks. Looks good!

operator/src/main/java/com/github/streamshub/console/api/v1alpha1/spec/containers/ContainerTemplate.java
```diff
@@ -0,0 +1,49 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonPropertyDescription;
+
+import io.fabric8.kubernetes.api.model.EnvVar;
+import io.fabric8.kubernetes.api.model.ResourceRequirements;
+import io.sundr.builder.annotations.Buildable;
+
+@Buildable
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class ContainerTemplate {
```
k-wall, line 14: Did you think of following the nomenclature  of https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-template-v1/#PodTemplateSpec?

In concrete terms, rename `ContainerTemplate` to `ContainerTemplateSpec` and have your ContainerSpec as a field within it.  At least that way you keep the door open to adding things like ContainerTemplateSpec.metadata.

k-wall (approved):
Looks reasonable, I left one suggestion to consider.

operator/src/main/java/com/github/streamshub/console/api/v1alpha1/spec/containers/ContainerTemplate.java
MikeEdgar: I think that makes sense. For an example, this means there is an added `spec` in the template, correct?

```yaml
spec:
  containers:
    api:
      spec:
        image: example.com/myorg/my-api-image:1.0
        resources:
          limits:
            cpu: 500m
            memory: 256Mi
        env:
        - name: SOME_ENVVAR
          value: "custom-value"
    ui:
      spec:
        image: example.com/myorg/my-ui-image:1.0
        resources:
          limits:
            cpu: 500m
            memory: 256Mi
        env:
        - name: SOME_ENVVAR_UI
          value: "custom-value"
```

MikeEdgar: Suggested applied - nested the container spec information within the container template, leaving a nice opening for `metadata` in the future if necessary.

PR #1246, opened 2024-11-26
Change image tags with digests in bundle metadata

jankalinic:
This PR introduces the use of SHA digests for built images in operator environment variables. Instead of relying on image tags, which can be unreliable for snapshot and latest tagged images, the operator now explicitly references UI and API images using their SHA digests.
The API and UI images will now always pull the correct version, ensuring consistency and avoiding issues caused by tag mismatches. 

This change was necessary because previously, using snapshot images with tags could result in the wrong image being pulled due to the console deployment's `IfNotPresent` image pull policy.

### DIFF
--- a/operator/bin/modify-bundle-metadata.sh
+++ b/operator/bin/modify-bundle-metadata.sh
@@ -17,27 +17,37 @@ do
   esac
 done
 
+api_name="console-api"
+ui_name="console-ui"
+operator_name="console-operator"
+
 echo "[INFO] Modify values and replace placeholders in ${CSV_FILE_PATH}"
 
 # Delete any namespaces set on the CSV deployment(s), possibly added by `quarkus.kubernetes.namespace`
 ${YQ} eval -o yaml -i 'del(.spec.install.spec.deployments[0].spec.template.metadata.namespace)' "${CSV_FILE_PATH}"
 
+# Get operator image name with tag
 yq_image_expression=".spec.install.spec.deployments[0] | (select (.name ==\"${ORIGINAL_OPERATOR_NAME}\")).spec.template.spec.containers[].image"
+operator_image_with_tag=$(${YQ} eval "${yq_image_expression}" "${CSV_FILE_PATH}")
+echo "[DEBUG] Original operator image name with tag = ${operator_image_with_tag}"
 
-full_image=$(${YQ} eval "${yq_image_expression}" "${CSV_FILE_PATH}")
-echo "[DEBUG] Original image name = ${full_image}"
+# Derive Registry and Tag for UI and API images
+image_registry=$(echo "${operator_image_with_tag}" | rev | cut -d':' -f2- | rev | sed "s|\/${operator_name}||")
+image_tag=$(echo "${operator_image_with_tag}" | cut -d':' -f2- )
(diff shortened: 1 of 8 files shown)


operator/bin/modify-bundle-metadata.sh
```diff
@@ -17,27 +17,37 @@ do
...
 ${YQ} eval -o yaml -i ".spec.relatedImages = null" "${CSV_FILE_PATH}"
 
-# Add operator image
-${YQ} eval -o yaml -i ".spec.relatedImages += [{\"name\": \"${OPERATOR_NAME}\", \"image\": \"${image_with_digest}\"}]" "${CSV_FILE_PATH}";
-echo "[DEBUG] Setting container image = ${image_with_digest}"
-${YQ} eval -o yaml -i ".metadata.annotations.containerImage = \"${image_with_digest}\"" "${CSV_FILE_PATH}"
+# Add operator image with digest to related images + replace operator image tag to digest
+echo "[DEBUG] Setting container image = ${operator_image_with_digest}"
+${YQ} eval -o yaml -i ".spec.relatedImages += [{\"name\": \"${OPERATOR_NAME}\", \"image\": \"${operator_image_with_digest}\"}]" "${CSV_FILE_PATH}";
+${YQ} eval -o yaml -i ".metadata.annotations.containerImage = \"${operator_image_with_digest}\"" "${CSV_FILE_PATH}"
```
MikeEdgar, line 50: I think we also need to use this image for `.spec.install.spec.deployments[0].spec.template.spec.containers[0].image`

MikeEdgar (commented):
Just one comment that I think needs an update for this PR. I also noticed an existing issue (not from this change) where the operator deployment and labels are named `streamshub-console-operator-v`, without the version. We can fix that in a separate PR.

operator/bin/modify-bundle-metadata.sh
MikeEdgar: @jankalinic I'll push this change to the PR as I know you'll be away for a while.

jankalinic: I think i played with that field, but when we set the`relatedImages` it suggests the digest for the operator deployment (i have always checked this value and it seemed correct)

jankalinic: Do you think that issue was solved by this PR? i think i've not seen that in the build job. (i might be mistaken and there might be a different cause issue)

MikeEdgar: I think the info in `relatedImages` is just metadata and won't be used to pull images, but I could be wrong. Either way, probably best to use the SHA in the deployment too, just to be consistent.

jankalinic: Sure, it's the bulletproof way.

MikeEdgar: Ok, it works, but I needed to update the runner images to ubuntu-24.04 for a newer version of `skopeo`. I'll merge this now.

MikeEdgar (approved):

operator/bin/modify-bundle-metadata.sh
jankalinic: Thank you for taking care of it!

PR #14, opened 2023-08-23
Add basic integration with console-api backend

MikeEdgar:
**What this PR does / why we need it**:

Integrate TS backend with [console-api](https://github.com/eyefloaters/console-api)

**Which issue(s) this PR fixes**

N/A

**Verification steps**

1. Deploy and run the console-api as described in the [README](https://github.com/eyefloaters/console-api/blob/65e4412d799b393034e495f9a6e725f93a07828e/README.md) (linked version is not yet on `main` branch).
2. Add `BACKEND_URL="http://127.0.0.1:8080"` to your `.env` file
3. Run `npm run dev`
4. Navigate to `/clusters` path in the UI

image

image

**Is there any work left / what's next** :

This is only a very rough example of integration with the console-api and is not expected to be merged as-is.

**Special notes for your reviewer**:

PR #1256, opened 2024-12-02
Enable ppc64le architecture support

Vaibhav-Nazare:
This PR introduces support for the ppc64le architecture, extending the project's usability across multiple architectures.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -67,7 +67,7 @@ jobs:
           mvn verify -P container-image -B --no-transfer-progress \
             -Dquarkus.kubernetes.namespace='$${NAMESPACE}' \
             -Dquarkus.package.write-transformed-bytecode-to-build-output=true \
-            -Dquarkus.docker.buildx.platform=linux/amd64,linux/arm64
+            -Dquarkus.docker.buildx.platform=linux/amd64,linux/arm64,linux/ppc64le
 
       - name: Modify CSV Annotation
         run: ./operator/bin/modify-bundle-metadata.sh "SKIP_RANGE=>=0.0.1 <${{ env.PROJECT_VERSION }}"
@@ -76,7 +76,7 @@ jobs:
         uses: docker/build-push-action@v6
         with:
           context: operator/target/bundle/console-operator/
-          platforms: linux/amd64,linux/arm64
+          platforms: linux/amd64,linux/arm64,linux/ppc64le
           provenance: false
           push: true
           file: operator/target/bundle/console-operator/bundle.Dockerfile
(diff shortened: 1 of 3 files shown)


jankalinic:
Thank you for the PR. I believe we should merge #1246 first and include these changes afterwards.

MikeEdgar (changes requested):
@Vaibhav-Nazare thanks for the PR. As Jan mentioned, there was another PR #1246 to make updates to these files, resulting in the conflict. Please rebase and update the same three files to specify the new platform in the `PLATFORMS` environment variable that is already present in each job.

jankalinic (approved):
I have not tried the changes, but overall looks good. Thank you!

jankalinic:
Wonder If 

>  #14 [linux/ppc64le 3/3] RUN ["/bin/opm", "serve", "/configs", "--cache-dir=/tmp/cache", "--cache-only"]
#14 0.773 time="2024-12-03T05:08:54Z" level=warning msg="unable to set termination log path" error="open /dev/termination-log: permission denied"

AND

>#15 [linux/arm64 3/3] RUN ["/bin/opm", "serve", "/configs", "--cache-dir=/tmp/cache", "--cache-only"]
#15 0.801 time="2024-12-03T05:08:54Z" level=fatal msg="could not start pprof endpoint: listen tcp 127.0.0.1:6060: bind: address already in use"
#15 ERROR: process "/bin/opm serve /configs --cache-dir=/tmp/cache --cache-only" did not complete successfully: exit code: 1

Is just build failure. Re-runing the job now. Please do not merge just yet, this needs to be fixed.

MikeEdgar:
I have definitely seen `listen tcp 127.0.0.1:6060: bind: address already in use` before. It looks like perhaps some race condition that only becomes more likely with 3 architectures being built.

Vaibhav-Nazare:
> I have definitely seen `listen tcp 127.0.0.1:6060: bind: address already in use` before. It looks like perhaps some race condition that only becomes more likely with 3 architectures being built.

I have seen this too on the first run that happened when I created the MR, but as I pulled in some changes and the build was triggered again,it got passed.

MikeEdgar:
> I have seen this too on the first run that happened when I created the MR, but as I pulled in some changes and the build was triggered again,it got passed.

Yeah, I think as long as the `RUN ["/bin/opm", "serve", "/configs", "--cache-dir=/tmp/cache", "--cache-only"]` command in the catalog Dockerfile does not run concurrently for any two architectures, it's fine. I think adding `network: none` [1] on the catalog build/push actions may help, but you don't need to include that here since it's an existing problem.

[1] https://docs.docker.com/reference/cli/docker/buildx/build/#network

PR #1260, opened 2024-12-03
Fix pagination issue in topic table

hemahg:
Fix the pagination issue in Topic table


MikeEdgar (approved):
Looks good to me, thanks @hemahg. I also found a bug in the API around how the pageNumber value is set which I added here.

Issue #1250, opened 2024-11-27
Partition 20+ cannot be selected in the message browser


MikeEdgar:
Select list is clipped with `message-browser.partitions_hidden`.

![image](https://github.com/user-attachments/assets/b17cddcf-6935-48be-b51b-f106cb7f3cd5)

PR #1262, opened 2024-12-03
fix:Partition 20+ cannot be selected in the message browser

hemahg:
fixes:Partition 20+ cannot be selected in the message [browser](https://github.com/streamshub/console/issues/1250)


MikeEdgar (approved):
LGTM

Issue #1265, opened 2024-12-03
NodesTable stories is using old value `Stable`


MikeEdgar:
The stories for the NodesTable should be updated to use the `Running` status name which maps to broker state `3`.

https://github.com/streamshub/console/blob/7b114d7a3caf33c74161fc98cc52bec29f17d96f/ui/app/%5Blocale%5D/(authorized)/kafka/%5BkafkaId%5D/nodes/nodesTable.stories.tsx#L4-L26

https://github.com/streamshub/console/blob/7b114d7a3caf33c74161fc98cc52bec29f17d96f/ui/app/%5Blocale%5D/(authorized)/kafka/%5BkafkaId%5D/nodes/page.tsx#L50-L73

PR #1268, opened 2024-12-03
fix node table story

hemahg:
fix node table story

related to [issue](https://github.com/streamshub/console/issues/1265)


MikeEdgar (approved):
Thanks Hema

Issue #1248, opened 2024-11-27
Allow use of self-signed certificates for an OIDC provider


MikeEdgar:
PR #1126 does not allow using self-signed certificates with the OIDC provider. We should allow trusted certificates to be configured, which requires the certificates to be passed to the UI server in PEM format, and the API server directly as input. The API server will then convert to a JKS written to temporary file to be passed to the OIDC tenant configuration. This step should not be required in the next Quarkus LTS.

| CR Input | UI | API |
| ------------- | ------------- | ------------- |
| PEM  | same as input | same as input |
| JKS  | Convert to PEM | same as input |
| P12  | Convert to PEM | same as input |

PR #1280, opened 2024-12-05
Configuration of trust stores for Prometheus and Apicurio clients

### LINKED ISSUE #1277: Support trusted certificate configuration for metrics sources and schema registries
Allow for a trusted certificate to be configured for metrics endpoints and schema registries

MikeEdgar:
Closes #1277 

--- 

Approach

This change allows for a trust store to be configured for metrics source (Prometheus) and schema registry (Apicurio) clients. The trust store may be set in the Console CR along with the URLs for each connection type.

For example:
```yaml
...
spec:
  metricsSources:
    - name: ocp-prometheus
      type: openshift-monitoring
      url: https://thanos-querier.mycluster.example.com
      truststore:
        type: PEM  # also supports JKS or PKCS12
        alias: "" # optional, only for JKS or PKCS12
        password: {} # optional, only for JKS or PKCS12. Same structure as `content` below
        content:
          # only one of `value` or `valueFrom` is used
          value: |
            -----BEGIN CERTIFICATE-----
            MIIFazCCA1OgAwIBAgIRAIIQz7DSQONZRGPgu2OCiwAwDQYJKoZIhvcNAQELBQAw
            ...
            emyPxgcYxn/eR44/KJ4EBs+lVDR3veyJm+kXQ99b21/+jh5Xos1AnX5iItreGCc=
            -----END CERTIFICATE-----
          # May use a ConfigMap or Secret key when no literal value is given
          valueFrom:
            configMapKeyRef: {}
            secretKeyRef: {}
...
```

The operator will then read the trust store content either directly from the CR for literal values or from the ConfigMap or Secret, copy to the Console secret being reconciled, and create a volume/mount and appropriate Quarkus environment vars for a TLS configuration bucket [1] specific to the connection. 

The clients in the API server will then obtain the trust store/SSL context from the TLS bucket named by convention and use for the connection to the remote service.

[1] https://quarkus.io/guides/tls-registry-reference#using-the-tls-registry

### DIFF
--- a/api/src/main/java/com/github/streamshub/console/api/service/MetricsService.java
+++ b/api/src/main/java/com/github/streamshub/console/api/service/MetricsService.java
@@ -38,6 +38,7 @@ import io.fabric8.kubernetes.client.KubernetesClient;
 import io.quarkus.tls.TlsConfiguration;
 import io.quarkus.tls.TlsConfigurationRegistry;
 
+import static com.github.streamshub.console.support.StringSupport.replaceNonAlphanumeric;
 import static java.util.stream.Collectors.groupingBy;
 import static java.util.stream.Collectors.mapping;
 import static java.util.stream.Collectors.toList;
@@ -52,7 +53,7 @@ public class MetricsService {
     Logger logger;
 
     @Inject
-    TlsConfigurationRegistry certificates;
+    TlsConfigurationRegistry tlsRegistry;
 
     @Inject
     KubernetesClient k8s;
@@ -91,15 +92,18 @@ public class MetricsService {
 
     public PrometheusAPI createClient(ConsoleConfig consoleConfig, KafkaClusterConfig clusterConfig) {
         PrometheusConfig prometheusConfig;
+        String sourceName = clusterConfig.getMetricsSource();
 
-        if (clusterConfig.getMetricsSource() != null) {
+        if (sourceName != null) {
             prometheusConfig = consoleConfig.getMetricsSources()
                     .stream()
-                    .filter(source -> source.getName().equals(clusterConfig.getMetricsSource()))
+                    .filter(source -> source.getName().equals(sourceName))
                     .findFirst()
                     .orElseThrow();
 
-            var trustStore = certificates.getDefault().map(TlsConfiguration::getTrustStore).orElse(null);
+            var trustStore = getTlsConfiguration(sourceName)
+                    .map(TlsConfiguration::getTrustStore)
+                    .orElse(null);
 
             RestClientBuilder builder = RestClientBuilder.newBuilder()
                     .baseUri(URI.create(prometheusConfig.getUrl()))
@@ -112,6 +116,12 @@ public class MetricsService {
         }
 
         return null;
+    }
+
+    Optional<TlsConfiguration> getTlsConfiguration(String sourceName) {
+        String dotSeparatedSource = "metrics.source." + replaceNonAlphanumeric(sourceName, '.');
+        String dashSeparatedSource = "metrics-source-" + replaceNonAlphanumeric(sourceName, '-');
+        return tlsRegistry.get(dotSeparatedSource).or(() -> tlsRegistry.get(dashSeparatedSource));
     }
 
     CompletionStage<Map<String, List<Metrics.ValueMetric>>> queryValues(String query) {
--- a/api/src/main/java/com/github/streamshub/console/api/support/serdes/ApicurioClient.java
+++ b/api/src/main/java/com/github/streamshub/console/api/support/serdes/ApicurioClient.java
@@ -1,0 +1,220 @@
+package com.github.streamshub.console.api.support.serdes;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.UncheckedIOException;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.util.Collections;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Supplier;
+
+import jakarta.enterprise.inject.spi.CDI;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.github.streamshub.console.config.SchemaRegistryConfig;
+
+import io.apicurio.registry.rest.client.impl.ErrorHandler;
+import io.apicurio.rest.client.auth.Auth;
+import io.apicurio.rest.client.error.RestClientErrorHandler;
+import io.apicurio.rest.client.request.Request;
+import io.apicurio.rest.client.spi.ApicurioHttpClient;
+import io.apicurio.rest.client.util.RegistryDateDeserializer;
+import io.apicurio.rest.client.util.UriUtil;
+import io.quarkus.tls.TlsConfiguration;
+import io.quarkus.tls.TlsConfigurationRegistry;
+
+import static com.github.streamshub.console.support.StringSupport.replaceNonAlphanumeric;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Apicurio client based on {@code io.apicurio.rest.client.JdkHttpClient}, but
+ * with awareness of Quarkus TLS registry configuration and removing other
+ * unused options for mTLS keystores, headers, etc.
+ *
+ * @see https://github.com/Apicurio/apicurio-common-rest-client/blob/0868773f61e33d40dcac88608aa111e26ab71bc7/rest-client-jdk/src/main/java/io/apicurio/rest/client/JdkHttpClient.java
+ */
+public class ApicurioClient implements ApicurioHttpClient {
+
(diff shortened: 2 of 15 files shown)


MikeEdgar:
@katheris , @k-wall this PR still needs some unit/integration tests, but I am curious to get your thoughts on the general approach/pattern to implement this.

api/src/main/java/com/github/streamshub/console/api/service/MetricsService.java
```diff
@@ -114,6 +118,12 @@ public PrometheusAPI createClient(ConsoleConfig consoleConfig, KafkaClusterConfi
         return null;
     }
 
+    Optional<TlsConfiguration> getTlsConfiguration(String sourceName) {
+        String dotSeparatedSource = "metrics.source." + replaceNonAlphanumeric(sourceName, '.');
+        String dashSeparatedSource = "metrics-source-" + replaceNonAlphanumeric(sourceName, '-');
```
MikeEdgar, line 123: The name may be slightly different (dots versus dashes) depending on whether the property was set via a system property or an environment variable.

operator/src/main/java/com/github/streamshub/console/dependents/ConsoleSecret.java
```diff
@@ -110,6 +130,119 @@ protected Secret desired(Console primary, Context<Console> context) {
...
+        @SuppressWarnings("unchecked")
+        List<VolumeMount> mounts = (List<VolumeMount>) deploymentResources.computeIfAbsent(VolumeMount.class, k -> new ArrayList<>());
+
+        mounts.add(new VolumeMountBuilder()
+                .withName(volumeName)
+                .withMountPath("/etc/ssl/" + sourcePrefix + sourceName + "." + typeCode)
+                .withSubPath(sourcePrefix + sourceName + "." + typeCode)
+                .build());
+
+        String configTemplate = "quarkus.tls.\"" + bucketPrefix + "-%s\".trust-store.%s.%s";
```
MikeEdgar, line 202: Ultimately, I'd like to avoid configuration of these buckets this way and have it supported via the console's YAML file, but that will be a bit too involved for the initial support I think.

k-wall:
>         alias: "" # optional, only for JKS or PKCS12

what's the idea here?  I'm familiar with alias on the keystore side -  to select a single key from potentially many in the store, but in the truststore side you normally trust all the certificates in the truststore.  In Java TrustManagerFactory accepts a truststore and gives you a TrustManager which trusts them all.  So, what's the role of the alias?

k-wall:
>         password: {} # optional, only for JKS or PKCS12. Same structure as `content` below

is this a ref to a secret?  you would normally try to avoid an inline secret.

MikeEdgar:
> > ```
> >     alias: "" # optional, only for JKS or PKCS12
> > ```
> 
> what's the idea here? I'm familiar with alias on the keystore side - to select a single key from potentially many in the store, but in the truststore side you normally trust all the certificates in the truststore. In Java TrustManagerFactory accepts a truststore and gives you a TrustManager which trusts them all. So, what's the role of the alias?

That's what I've typically seen also, but the TLS registry supports configuration of an `alias` for JKS and P12 [1]. I figured better to align to the target on it.

[1] https://quarkus.io/guides/tls-registry-reference#trust-stores

MikeEdgar:
> > ```
> >     password: {} # optional, only for JKS or PKCS12. Same structure as `content` below
> > ```
> 
> is this a ref to a secret? you would normally try to avoid an inline secret.

Yes, it can be. The field allows for `value` or `valueFrom` like the truststore `content`. I kept the structure the same to allow security, but also convenience for development and testing, etc.

api/src/main/java/com/github/streamshub/console/api/support/serdes/ApicurioClient.java
```diff
@@ -0,0 +1,220 @@
...
+    @Override
+    public <T> T sendRequest(Request<T> request) {
+        try {
+            requireNonNull(request.getOperation(), "Request operation cannot be null");
+            requireNonNull(request.getResponseType(), "Response type cannot be null");
+
+            final HttpRequest.Builder requestBuilder = HttpRequest.newBuilder()
+                    .uri(UriUtil.buildURI(endpoint + request.getRequestPath(), request.getQueryParams(), request.getPathParams()));
+
+            //Add current request headers
```
k-wall, line 101: Oh my!

api/src/main/java/com/github/streamshub/console/api/support/serdes/ApicurioClient.java
```diff
@@ -0,0 +1,220 @@
...
+import io.apicurio.rest.client.spi.ApicurioHttpClient;
+import io.apicurio.rest.client.util.RegistryDateDeserializer;
+import io.apicurio.rest.client.util.UriUtil;
+import io.quarkus.tls.TlsConfiguration;
+import io.quarkus.tls.TlsConfigurationRegistry;
+
+import static com.github.streamshub.console.support.StringSupport.replaceNonAlphanumeric;
+import static java.util.Objects.requireNonNull;
+
+/**
```
k-wall, line 37: Do we have a path to removing this code from the Console?

operator/src/main/java/com/github/streamshub/console/dependents/ConsoleDeployment.java
```diff
@@ -58,6 +64,11 @@ protected Deployment desired(Console primary, Context<Console> context) {
         String imageAPI = imagesSpec.map(Images::getApi).orElse(defaultAPIImage);
         String imageUI = imagesSpec.map(Images::getUi).orElse(defaultUIImage);
 
+        var envVars = new ArrayList<>(coalesce(primary.getSpec().getEnv(), Collections::emptyList));
```
k-wall, line 67: I don't understand why `coalesce` is called coalesce.  it seems to be `valueOrDefault()`

operator/src/main/java/com/github/streamshub/console/dependents/ConsoleDeployment.java
```diff
@@ -82,9 +93,13 @@ protected Deployment desired(Console primary, Context<Console> context) {
                                 .withSecretName(configSecretName)
                             .endSecret()
                         .endVolume()
+                        .addAllToVolumes(getResourcesByType(trustResources, Volume.class))
                         .editMatchingContainer(c -> "console-api".equals(c.getName()))
                             .withImage(imageAPI)
-                            .addAllToEnv(coalesce(primary.getSpec().getEnv(), Collections::emptyList))
+                            .addAllToVolumeMounts(getResourcesByType(trustResources, VolumeMount.class))
+                            // Remove first to avoid duplicates
+                            .removeMatchingFromEnv(env -> envVars.stream().anyMatch(e -> env.getName().equals(e.getName())))
```
k-wall, line 101: the approach looks worrying.   doesn't this potentially overwrite env vars coming from the `console.deployment.yaml` with values provided by the user from the CR.

operator/src/main/java/com/github/streamshub/console/dependents/ConsoleDeployment.java
MikeEdgar: This is my SQL history coming in :laughing: ... it's also named something similar in JavaScript [1]. Functionally it's `valueOrDefault`, but less verbose.

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

api/src/main/java/com/github/streamshub/console/api/support/serdes/ApicurioClient.java
MikeEdgar: I haven't given too much thought yet, other than to think that this kind of support should be in the Quarkus extension [1] somehow.

[1] https://github.com/quarkusio/quarkus/tree/main/extensions/schema-registry/apicurio

operator/src/main/java/com/github/streamshub/console/dependents/ConsoleDeployment.java
MikeEdgar: That's a good point, the ENVs in the operator should always be present. If user attempts to override it results in a deployment error because they are duplicates. That is probably the correct behavior, but it happened to me trying to debug this and I added this remove line in haste. I'll go ahead and take it out.

api/src/main/java/com/github/streamshub/console/api/support/serdes/ApicurioClient.java
k-wall: Yes, I think it is worth separately opening a RFE/putting up a PR to get this code back where it belongs.  Although maybe it is worth holding off until the dust from Apicurio 3.0 has settled.

k-wall:
On the whole, the approach looks good.  I'm concerned about some of the coding in the ApicurioClient but will be content to overlook providing there's a plan to address the debt.

api/src/main/java/com/github/streamshub/console/api/support/serdes/ApicurioClient.java
MikeEdgar: I've opened https://github.com/quarkusio/quarkus/issues/45069 to track this.

MikeEdgar:
@k-wall I'll go ahead and merge this if no objections. The config option in the CR will be used also for #1248 to support a TLS truststore for the OIDC provider (building on #1126 once that's merged)

k-wall (approved):

PR #1303, opened 2024-12-12
fix the icon on rebalance page

hemahg:
Update the icons for "Not ready" and "stopped" state in Rebalancing table


MikeEdgar (approved):
LGTM

PR #1314, opened 2024-12-17
Security via operator, OIDC trusted certificates

### LINKED ISSUE #1247: Support security configuration via the operator
PR #1126 does not yet include support for security configuration via the operator. This should be supported using the `Console` CR.

### LINKED ISSUE #1248: Allow use of self-signed certificates for an OIDC provider

MikeEdgar:
1. Enable security to be configured via operator/Console CR.  This allows an OIDC provider, security subjects (users or groups), and roles with access rules to be given in the CR using the same structure as the API server's YAML configuration file.
2. Support the use of a truststore for the OIDC identity provider. The truststore must be passed to both the API and the UI in potentially different formats. The API supports and of JKS, P12, or PEM truststores, but the UI must have it provided as a PEM chain.
3. Modify API integration tests to use TLS with Keycloak instance.
4. Move configuration mapping in operator code from `ConsoleSecret` dependent resource to new `ConfigurationProcessor` "virtual" resource upon which all other dependent resources depend on. This provides a single point of validation at the start of each reconcile process.
5. Improve deployment ready condition in operator, taking behavior from `kubectl`'s deployment rollout status check.

Closes #1247 
Closes #1248 

Issue #1335 will further refine how truststores are configured in the API server by avoiding the use of `quarkus.tls.*` configuration properties and supporting configuration in the console's own YAML file.


api/src/main/java/com/github/streamshub/console/api/security/ConsoleAuthenticationMechanism.java
```diff
@@ -102,6 +102,10 @@ boolean oidcEnabled() {
 
     @Override
     public Uni<SecurityIdentity> authenticate(RoutingContext context, IdentityProviderManager identityProviderManager) {
+        if (!context.normalizedPath().startsWith("/api")) {
+            return Uni.createFrom().nullItem();
+        }
```
MikeEdgar, line 107: No authentication for non-API paths. E.g. `/metrics`, `/health`, and `/openapi`

api/src/main/java/com/github/streamshub/console/api/security/OidcTenantConfigResolver.java
```diff
@@ -40,6 +63,44 @@ void initialize() {
...
+        );
+    }
+
+    Optional<TlsConfiguration> getTlsConfiguration() {
+        String dotSeparatedSource = "oidc.provider.trust";
+        String dashSeparatedSource = "oidc-provider-trust";
+        return tlsRegistry.get(dotSeparatedSource).or(() -> tlsRegistry.get(dashSeparatedSource));
+    }
+
+    void configureTruststore(KeyStore truststore) {
```
MikeEdgar, line 79: The OIDC subsystem takes the path to a truststore, so we need to write the one from the TLS registry to a working file to provide to OIDC. This should no longer be necessary in the next Quarkus LTS where OIDC is aware of the TLS registry.

api/src/main/resources/application.properties
```diff
@@ -68,6 +68,7 @@ quarkus.arc.exclude-types=io.apicurio.registry.rest.JacksonDateTimeCustomizer
 quarkus.index-dependency.strimzi-api.group-id=io.strimzi
 quarkus.index-dependency.strimzi-api.artifact-id=api
 
+console.work-path=${java.io.tmpdir}
```
MikeEdgar, line 71: The operator overrides this to use an emptydir volume in Kube, maintaining a read-only root container filesystem.

operator/src/main/java/com/github/streamshub/console/dependents/ConsoleDeployment.java
```diff
@@ -130,4 +136,8 @@ <R extends KubernetesResource> List<R> getResourcesByType(
             Class<R> key) {
         return (List<R>) resources.getOrDefault(key, Collections.emptyList());
     }
+
+    private String pullPolicy(String image) {
+        return image.contains("sha256:") ? "IfNotPresent" : "Always";
+    }
```
MikeEdgar, line 142: The idea here is there is no point re-pulling an image that uses a SHA, but a tagged image may have changed and should be re-pulled. Default configuration in the released operator will use image SHAs.

api/src/main/java/com/github/streamshub/console/api/security/ConsoleAuthenticationMechanism.java
k-wall: I think security best practice would be to flip it the other way around.   i.e list of the paths that allowed without auth and let everything else require it.

api/src/main/java/com/github/streamshub/console/api/security/OidcTenantConfigResolver.java
k-wall: Might be worth adding that as comment.

api/src/main/java/com/github/streamshub/console/api/security/OidcTenantConfigResolver.java
```diff
@@ -40,6 +63,44 @@ void initialize() {
...
+        File workDir = new File(workPath);
+        File truststoreFile;
+
+        try {
+            truststoreFile = File.createTempFile("oidc-provider-trust", "." + truststore.getType(), workDir);
```
k-wall, line 84: With a least privilege hat on, consider using the 3 arg form of File.createTempFile and pass attrs as
PosixFilePermissions.asFileAttribute(PosixFilePermissions.fromString("rw-------")

api/src/test/java/com/github/streamshub/console/kafka/systemtest/deployment/KeycloakResourceManager.java
```diff
@@ -29,26 +33,56 @@ public Map<String, String> start() {
             throw new UncheckedIOException(ioe);
         }
 
+        int port = 8443;
+        TlsHelper tls = TlsHelper.newInstance();
+        String keystorePath = "/opt/keycloak/keystore.p12";
+
         keycloak = new GenericContainer<>("quay.io/keycloak/keycloak:26.0")
```
k-wall, line 40: I'd extract a constant for the image so that it is easier to find when it is time to upgrade.

api/src/main/java/com/github/streamshub/console/api/security/OidcTenantConfigResolver.java
MikeEdgar: Good idea, I'll add that.

k-wall:
Could you expand on the "Enable security to be configured via operator/Console CR" bullet in the PR's description.  I'm lacking some context.

api/src/main/java/com/github/streamshub/console/api/security/OidcTenantConfigResolver.java
```diff
@@ -40,6 +63,44 @@ void initialize() {
...
+            truststoreFile.deleteOnExit();
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+
+        String secret = UUID.randomUUID().toString();
+
+        try (OutputStream out = new FileOutputStream(truststoreFile)) {
+            truststore.store(out, secret.toCharArray());
```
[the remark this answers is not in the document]
MikeEdgar: The secret is only passed to `store`, it shouldn't end up in the resulting file.

api/src/test/java/com/github/streamshub/console/kafka/systemtest/deployment/KeycloakResourceManager.java
MikeEdgar: I've been thinking about using a property in the POM, along with a dependency to a Keycloak artifact so that dependabot will bump this.

api/src/main/java/com/github/streamshub/console/api/security/ConsoleAuthenticationMechanism.java
MikeEdgar: Done - replaced the `!context.normalizedPath().startsWith("/api")` with a check against a predefined set of paths that should not be authenticated.

api/src/main/java/com/github/streamshub/console/api/security/OidcTenantConfigResolver.java
MikeEdgar: Moved the original comment to the method's JavaDoc
> The OIDC subsystem takes the path to a truststore, so we need to write the one from the TLS registry to a working file to provide to OIDC. This should no longer be necessary in the next Quarkus LTS where OIDC is aware of the TLS registry.

MikeEdgar:
> Could you expand on the "Enable security to be configured via operator/Console CR" bullet in the PR's description. I'm lacking some context.

Updated. Let me know if you have any questions on it.

> 1. Enable security to be configured via operator/Console CR.  This allows an OIDC provider, security subjects (users or groups), and roles with access rules to be given in the CR using the same structure as the API server's YAML configuration file.

api/src/test/java/com/github/streamshub/console/kafka/systemtest/deployment/KeycloakResourceManager.java
k-wall: (aside:   I'd been looking for a good way to use dependabot for image updates in [Kroxylicious](https://github.com/kroxylicious/kroxylicious/issues/1179))

api/src/main/java/com/github/streamshub/console/api/security/OidcTenantConfigResolver.java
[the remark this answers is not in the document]
k-wall: gah, sorry, I completely misread the code.

operator/src/main/java/com/github/streamshub/console/api/v1alpha1/spec/security/Audit.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    public enum Decision {
+        ALLOWED,
+        DENIED,
+        ALL;
+
+        @JsonCreator
+        public static Decision forValue(String value) {
+            if ("*".equals(value)) {
```
k-wall, line 31: Do we need to guard null?

operator/src/main/java/com/github/streamshub/console/api/v1alpha1/spec/security/Audit.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.Locale;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+
+import io.fabric8.generator.annotation.Required;
+import io.sundr.builder.annotations.Buildable;
+
+@Buildable(editableEnabled = false)
+public class Audit extends Rule {
```
k-wall, line 11: Naming. Is this an AuditingRule i.e. a rule that audits its decision?

operator/src/main/java/com/github/streamshub/console/api/v1alpha1/status/Condition.java
```diff
@@ -61,4 +67,70 @@ public String getMessage() {
...
+        this.active = active;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(message, reason, status, type);
+    }
+
+    /**
+     * For the purposes of equality, we do not consider the lastTransitionTime.
```
k-wall, line 85: Is active intentionally omitted from the equality too?

operator/src/main/java/com/github/streamshub/console/dependents/ConfigurationProcessor.java
```diff
@@ -0,0 +1,744 @@
...
+     * method will add to the context the resources to be added to the console deployment to
+     * access the secret entries.
+     */
+    private void buildTrustStores(Console primary, Context<Console> context, Map<String, String> data) {
+        var processor = new TrustStoreProcessor(primary, context, data);
+        processor.process();
+    }
+
+    private static String base64String(int length) {
+        ByteArrayOutputStream buffer = new ByteArrayOutputStream();
```
k-wall, line 430: nit - method name doesn't reflect the implementation.   (generateRandomSecret)
nit - use try with resources.

operator/src/test/java/com/github/streamshub/console/ConsoleReconcilerTestBase.java
```diff
@@ -0,0 +1,317 @@
...
+import io.javaoperatorsdk.operator.Operator;
+import io.strimzi.api.kafka.Crds;
+import io.strimzi.api.kafka.model.kafka.Kafka;
+import io.strimzi.api.kafka.model.kafka.KafkaBuilder;
+import io.strimzi.api.kafka.model.kafka.listener.KafkaListenerAuthenticationScramSha512;
+import io.strimzi.api.kafka.model.kafka.listener.KafkaListenerType;
+import io.strimzi.api.kafka.model.user.KafkaUser;
+
+import static org.awaitility.Awaitility.await;
+import static org.junit.Assert.assertNull;
```
k-wall, line 44: nit: mixing jupiter and pre-jupiter.

k-wall:
It is a big PR, so reviewing has been tough going.  LGTM.
In future, if it is possible to decompose into smaller PRs reviews will be easier.

k-wall (approved):

operator/src/main/java/com/github/streamshub/console/api/v1alpha1/status/Condition.java
MikeEdgar: Yes, `active` is only meant as a marker for Conditions that still apply at the end of the reconcile process. This helps maintain an accurate `lastTransitionTime` for a given condition. I'll update the comment that it's excluded as well.

operator/src/main/java/com/github/streamshub/console/api/v1alpha1/spec/security/Audit.java
MikeEdgar: This Audit class is used to enable audit logging for authorization decisions for resources that match the rule criteria in the parent class. Renaming to `AuditRule` would make sense.

operator/src/main/java/com/github/streamshub/console/api/v1alpha1/spec/security/Audit.java
MikeEdgar: > Do we need to guard null?

Kube rejects the CR with:
> Error "Required value" for field "spec.security.audit[0].decision".

Similarly, attempts to use a null for an entry in the `privileges` array results in:
> Error "Invalid value: "null": spec.security.audit[0].privileges[1] in body must be of type string: "null"" for field "spec.security.audit[0].privileges[1]".

k-wall:
Thanks for the update, lgtm

api/src/test/java/com/github/streamshub/console/kafka/systemtest/deployment/KeycloakResourceManager.java
MikeEdgar: Opened https://github.com/streamshub/console/issues/1349 to track this for an upcoming fix.

PR #1319, opened 2024-12-19
Bug fix in reset offset page

hemahg:
1. **Reset Offset Button Visibility**: Fix an issue where the "Reset Offset" button remains visible while already on the reset offset form. This occurs when navigating from the "Reset Offset" button on the consumer group page (but not when accessed via the list kebab menu).
2. **Remove Description Text in Kebab Menu**: Remove the description text in the kebab menu for consumer groups where the status is empty.
3. **Disable Dryrun Button Until Offset is Selected**: Ensure the "Dryrun" button remains disabled until an offset value is selected.
4. **Restrict Custom Input to Non-Negative Values**: Prevent negative numbers from being entered in the "Custom Offset" input field.

### DIFF
--- a/ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/[groupId]/reset-offset/ResetOffset.tsx
+++ b/ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/[groupId]/reset-offset/ResetOffset.tsx
@@ -83,6 +83,20 @@ export function ResetOffset({
   onOffsetSelect: (value: OffsetValue) => void;
 }) {
   const t = useTranslations("ConsumerGroupsTable");
+
+  const isDisabled =
+    (selectTopic === "allTopics" ||
+      (selectTopic === "selectedTopic" &&
+        offset.topicName &&
+        (selectPartition === "allPartitions" ||
+          (selectPartition === "selectedPartition" &&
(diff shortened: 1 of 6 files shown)


ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/[groupId]/reset-offset/ResetOffset.tsx
```diff
@@ -219,11 +238,15 @@ export function ResetOffset({
...
-                isDisabled={isLoading}
+                isDisabled={isLoading || !isDisabled}
               >
                 {t("save")}
               </Button>
-              <DryrunSelect openDryrun={openDryrun} cliCommand={cliCommand} />
+              <DryrunSelect
+                openDryrun={openDryrun}
+                cliCommand={cliCommand}
+                isDisabled={!isDisabled}
```
Comment (MikeEdgar) [inline ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/[groupId]/reset-offset/ResetOffset.tsx:248]: This seems somewhat confusing. The `DryrunSelect` is disabled when the criteria here are "not disabled". Should the variable in this file be named differently, or am I missing something?

MikeEdgar (approved):
Logic and functionality look good to me. I just have one question on the naming of a variable that I don't follow.

MikeEdgar (approved):
Thanks @hemahg

PR #1336, opened 2025-01-02
fix: resolve several next-intl deprecation warnings

MikeEdgar:


ui/i18n/routing.ts
```diff
@@ -14,5 +14,13 @@ export const routing = defineRouting({
...
+let nav = createNavigation(routing);
+
+export const { Link, redirect, usePathname, useRouter } = {
+    Link: nav.Link,
+    redirect: (path: string) => {
+        nav.redirect({ href: path, locale: "en" });
+    },
+    usePathname: nav.usePathname,
+    useRouter: nav.useRouter,
+};
```
MikeEdgar, line 26: The `redirect` function requires the locale on each use now. This is a cheat to use the default since we're only supporting `en` at the moment.

hemahg (approved):

Issue #1286, opened 2024-12-05
Buggy row display in message browser


MikeEdgar:
The display of the rows in the message browser seems to be buggy, with sometimes a blank row appearing and some rows shifting in the row above/below. This occurs for me in both Firefox and Chrome.

![image](https://github.com/user-attachments/assets/405c4f23-811a-4c03-87c8-0958461922a4)

PR #1342, opened 2025-01-07
fix buggy row in message browser

hemahg:
fix the buggy row in the message browser

related to [issue1286](https://github.com/streamshub/console/issues/1286)


MikeEdgar (commented):
It's looking better, but I do still see some blank rows when compressing the table with schema links.
image

MikeEdgar (approved):
It appears to be an issue on my side only. Let's merge it and address later if it is still an issue for others.

PR #1346, opened 2025-01-08
Bug fix in consumer group reset offset

hemahg:
Fix redundant "0" displayed in partition dropdown when "A selected partition" is selected


MikeEdgar (approved):
LGTM. I never noticed this :smile:

PR #1350, opened 2025-01-09
Use dependabot for test image updates, rm obsolete/cleanup test code

### LINKED ISSUE #1349: Ensure image references in tests are properly updated by dependabot
Container images used in several tests are currently hard-coded in Java classes. We should have these externalized somehow so that dependabot can update them to ensure the tests stay on current versions of the containers.

Idea 1:
Use a Maven property and a "dummy" dependency of the dependency that dependabot Maven can track for updates. This has the downside of adding a fake dependency and also tracking a version string that may differ from the actual container image tag.

Idea 2:
Add an empty Dockerfile for each image in the test resources that may be loaded and the image reference parsed for use in the tests. The Dockerfiles would be updated automatically by dependabot Docker support. This seems like a better choice than "idea 1".

MikeEdgar:
Update API tests to fetch container image references from Dockerfiles that will be tracked by dependabot for updates. I also took this opportunity to prune out a bunch of obsolete/unused test code that have been around from the Kafka Admin API project that the API was forked from.

Closes #1349


MikeEdgar:
I will go ahead and merge this change, test code only.

jankalinic (commented):
LGTM

PR #1351, opened 2025-01-10
Hide custom offset select when all topics is selected in Consumer group reset offset 

hemahg:
Hide custom offset select when all topics are selected in the Consumer group reset offset


ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/[groupId]/reset-offset/ResetOffset.tsx
```diff
@@ -179,21 +179,32 @@ export function ResetOffset({
...
-                    { value: "custom", label: t("offset.custom") },
-                    { value: "latest", label: t("offset.latest") },
-                    { value: "earliest", label: t("offset.earliest") },
-                    {
-                      value: "specificDateTime",
-                      label: t("offset.specific_date_time"),
-                    },
-                  ]}
+                  options={
+                    selectTopic === "allTopics"
```
Comment (MikeEdgar) [inline ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/[groupId]/reset-offset/ResetOffset.tsx:183]: I think the specific offset should also be disabled for "All Partitions".

```suggestion
                    selectTopic === "allTopics" || selectPartition === "allPartitions"
```

Side note: reset/dry run button remain disabled when entering an offset of `0`. Probably due to `0` being "falsy" when setting the `isEnabled` boolean.

MikeEdgar (changes requested):
Works well for all topics, but I think we need it for all partitions too.

ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/[groupId]/reset-offset/ResetOffset.tsx
Comment (hemahg) [reply ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/[groupId]/reset-offset/ResetOffset.tsx]: fixed both the issue

MikeEdgar (approved):
Looks good, thanks @hemahg

PR #1383, opened 2025-01-17
integration: add UI build cache, split Maven tests to separate job

MikeEdgar:
An attempt to improve the CI build time

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -46,6 +46,18 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-maven-
 
+      - name: Cache Next.js UI Build
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.npm
+            ${{ github.workspace }}/.next/cache
+          # Generate a new cache whenever packages or source files change.
+          key: ${{ runner.os }}-nextjs-${{ hashFiles('**/package-lock.json') }}-${{ hashFiles('**/*.js', '**/*.jsx', '**/*.ts', '**/*.tsx') }}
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yml
```diff
@@ -77,26 +88,15 @@ jobs:
...
-          #
-          # See: https://quarkus.io/guides/tests-with-coverage#coverage-for-integration-tests
-          #
           export QUARKUS_CONTAINER_IMAGE_REGISTRY="localhost:5000"
           export QUARKUS_CONTAINER_IMAGE_PUSH=true
           export QUARKUS_CONTAINER_IMAGE_TAG="${{ env.PROJECT_VERSION }}"
           export QUARKUS_KUBERNETES_VERSION="${{ env.PROJECT_VERSION }}"
-          mvn verify -P container-image -B --no-transfer-progress \
+
+          mvn verify -P container-image -B --no-transfer-progress -DskipTests \
```
jankalinic, line 98: Do we want to skip tests here? I know it saves a lot of time and i see you have added parallel test-java to workflow, so it might be alright, but will that use more GH storage if the build is okay, but tests fail? I think the saving images task was skiped when the test failed so it did not produce any artifacts. If this is not a problem, then i think we can merge. Thanks!

MikeEdgar: @jankalinic I think it's OK. The storage is temporary and Github deletes it after a while anyway. We could also add the `test-java` job to the `needs` of the Playwright step, but that would still use the storage to upload the images.

jankalinic (approved):

PR #1393, opened 2025-01-21
Move schema link to message details drawer

hemahg:
- Move the schema link to the message details drawer
Screenshot 2025-01-21 at 10 54 27 PM

- Improve how the schema is displayed in the new browser tab
Screenshot 2025-01-21 at 10 54 37 PM


ui/middleware.ts
```diff
@@ -43,22 +43,33 @@ const protectedPathnameRegex = RegExp(
...
   const requestPath = req.nextUrl.pathname;
-  const isPublicPage = !oidcEnabled && publicPathnameRegex.test(requestPath);
-  const isProtectedPage = oidcEnabled || protectedPathnameRegex.test(requestPath);
+  const isSchemaPublic =
+    requestPath === "/schema" &&
+    searchParams.has("content") &&
+    searchParams.has("schemaname");
+
+  const isPublicPage =
+    (!oidcEnabled && publicPathnameRegex.test(requestPath)) || isSchemaPublic;
```
MikeEdgar, line 65: Making schemas public won't work without API changes. When OIDC is enabled, the entire `/api` path will be protected so these requests will return a 401. Can the schemas still be under `(authorized)` somehow? Is it conflicting with some other path?

hemahg: I understand the concern about the /api path being protected when OIDC is enabled. However, I am passing anonymous: true in the headers by using await getHeaders(true). This should allow the request to bypass authentication, similar to how Kafka logging is handled.

hemahg: Ah okay, I see the problem

hemahg: I have updated the PR to support OIDC, hope this will fix the issue

MikeEdgar (commented):
This works now, but I wonder what is the reason for bypassing security and using the `(public)` directory for schema. 

Something I noticed on the API logs while testing this is that when I select a message row in the listing and the drawer opens, the backend receives 8 identical requests for the schema. I don't see anything obvious to indicate why it's calling so many times thought.

hemahg:
The schema doesn't require any of the navigation layout components (side navigation, logo, header, etc.) that are applied to all the paths under the Kafka folder. Therefore, I bypassed these components and placed the schema in the public folder to simplify the design and make it accessible without unnecessary overhead.

MikeEdgar:
Ok, that makes sense. So the requests to the backend are not actually unauthenticated/public in the case of OIDC.

hemahg:
yes, backend still remains authenticated

MikeEdgar (approved):

PR #1394, opened 2025-01-22
[Operator] Modify CSV to add support for OperatorHub on arm64,amd64,ppc64le,s390x

jankalinic:
This PR makes Console Operator to be visible in OperatorHub. Previously this was not possible. Even though the Operator could be installed via subscription from CLI and was fully functional, it did not show up in the Openshift OperatorHub, making it harder for users to install using UI.

### DIFF
--- a/operator/bin/modify-bundle-metadata.sh
+++ b/operator/bin/modify-bundle-metadata.sh
@@ -111,6 +111,11 @@ ${YQ} eval -o yaml -i '.spec.relatedImages += [{
   "image": "'${ui_image_with_digest}'"
 }]' "${CSV_FILE_PATH}"
 
+${YQ} eval -o yaml -i ".metadata.labels[\"operatorframework.io/arch.amd64\"] = \"supported\"" "${CSV_FILE_PATH}"
+${YQ} eval -o yaml -i ".metadata.labels[\"operatorframework.io/arch.arm64\"] = \"supported\"" "${CSV_FILE_PATH}"
+${YQ} eval -o yaml -i ".metadata.labels[\"operatorframework.io/arch.ppc64le\"] = \"supported\"" "${CSV_FILE_PATH}"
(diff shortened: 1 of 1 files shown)


operator/bin/modify-bundle-metadata.sh
```diff
@@ -111,6 +111,11 @@ ${YQ} eval -o yaml -i '.spec.relatedImages += [{
   "image": "'${ui_image_with_digest}'"
 }]' "${CSV_FILE_PATH}"
 
+${YQ} eval -o yaml -i ".metadata.labels[\"operatorframework.io/arch.amd64\"] = \"supported\"" "${CSV_FILE_PATH}"
+${YQ} eval -o yaml -i ".metadata.labels[\"operatorframework.io/arch.arm64\"] = \"supported\"" "${CSV_FILE_PATH}"
+${YQ} eval -o yaml -i ".metadata.labels[\"operatorframework.io/arch.ppc64le\"] = \"supported\"" "${CSV_FILE_PATH}"
+${YQ} eval -o yaml -i ".metadata.labels[\"operatorframework.io/arch.s390x\"] = \"supported\"" "${CSV_FILE_PATH}"
+
```
MikeEdgar, line 118: Just a suggestion since there should be a `PLATFORMS` variable defined when this is running. I also recently found that since the bundle image itself is not platform-specific, the `Build Operator Bundle Image` step should not even need the `platforms: ${{ env.PLATFORMS }}` property. Maybe we can eliminate that in the future too.

```suggestion
for full_arch in ${PLATFORMS//,/ } ; do 
  os_arch=($(echo ${full_arch} | tr '/' '\n'))
  arch=${os_arch[1]}
  ${YQ} eval -o yaml -i ".metadata.labels[\"operatorframework.io/arch.${arch}\"] = \"supported\"" "${CSV_FILE_PATH}"
done
```

MikeEdgar (approved):

MikeEdgar (approved):

MikeEdgar:
@jankalinic this is OK to merge right?

PR #1402, opened 2025-01-24
Fix consumer group filter not working issue

hemahg:
Fix consumer group filter not working issue


ui/api/consumerGroups/actions.ts
```diff
@@ -36,22 +43,21 @@ export async function getConsumerGroups(
     filterUndefinedFromObj({
       "fields[consumerGroups]":
         params.fields ?? "state,simpleConsumerGroup,members,offsets",
-      "filter[id]": params.id ? `eq,${params.id}` : undefined,
-      // TODO: pass filter from UI
-      "filter[state]":
-        params.state && params.state.length > 0
-          ? `in,${params.state.join(",")}`
-          : undefined,
+      "filter[id]": filterEq(params.id),
```
MikeEdgar, line 46:
```suggestion
      "filter[id]": filterLike(params.id),
```

ui/api/consumerGroups/actions.ts
```diff
@@ -1,5 +1,12 @@
 "use server";
-import { fetchData, patchData, sortParam, ApiResponse } from "@/api/api";
+import {
+  fetchData,
+  patchData,
+  sortParam,
+  ApiResponse,
+  filterEq,
```
MikeEdgar, line 7:
```suggestion
  filterLike,
```

MikeEdgar (commented):
I tried this locally and confirmed it's working now. The only issue might be the removal of the auto-refresh. Is it broken once filters are introduced?

MikeEdgar (approved):
After discussion, we decided it is a good idea to remove the auto-refresh on the groups because (1) nowhere else in the application behaves this way and (2) when there are a many consumer groups and a user leaves a browser open on this page, it could make an excessive number of requests to Kafka to retrieve that data needed for the listing.

ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/ConsumerGroupsTable.tsx
```diff
@@ -76,31 +75,17 @@ export function ConsumerGroupsTable({
...
-        if (consumerGroups) {
-          setConsumerGroups(consumerGroups);
-        }
-      }, 5000);
-    }
-    return () => clearInterval(interval);
-  }, [refresh]);
   return (
     <TableView
-      itemCount={consumerGroups?.length}
```
Comment (MikeEdgar) [inline ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/ConsumerGroupsTable.tsx:103]: @hemahg , I didn't notice earlier, but I think this line change disables pagination.

Comment (hemahg) [reply ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/ConsumerGroupsTable.tsx:103]: pagination will be applied if the number of records is more than 20

Comment (MikeEdgar) [reply ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/ConsumerGroupsTable.tsx:103]: Ok, got it. Somehow it's not working with the storybook for this page. See the changes in chromatic for the build: https://www.chromatic.com/build?appId=66cf23a2f89a8283f775d05b&number=328

PR #1404, opened 2025-01-24
fix: default to local platform, use docker additional-args Quarkus flag

MikeEdgar:
- Default `PLATFORMS` in Makefile to just the current native platform.
- Use `QUARKUS_DOCKER_ADDITIONAL_ARGS` instead of `QUARKUS_DOCKER_BUILDX_PLATFORM` (does not work with podman)


MikeEdgar:
@jankalinic I think this should also work on Mac, but when you have time can you check? This is low priority.

jankalinic:
Sure, i can put it on my todo. Thanks for letting me know about this PR.

jankalinic:
But if you are okay with quick check it returns `linux/arm64` on MacPro M2

PR #1417, opened 2025-01-28
Create separate header for reset offset and dry-run page

hemahg:
- Create a separate route for Reset offset and dry-run page.
- Create a separate header for reset offset and dry-run page.

ResetOffset Page
Screenshot 2025-01-28 at 2 38 40 PM

Dryrun Page
Screenshot 2025-01-28 at 2 38 18 PM


riccardo-forina (approved):
Changes look good. I couldn’t try this, is there a way to go back to the previous page from the dry run one?

hemahg:
Yes, the "Back to edit offset" button in the dryrun, and also breadcrumb will take to the reset offset page

MikeEdgar (approved):

PR #1426, opened 2025-01-31
Add proper error message when incorrect custom offset value is entered

hemahg:
Add proper error message when incorrect custom offset value is entered

### DIFF
--- a/ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/[groupId]/reset-offset/ResetOffset.tsx
+++ b/ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/[groupId]/reset-offset/ResetOffset.tsx
@@ -1,12 +1,7 @@
 import {
-  Divider,
   Panel,
-  PanelHeader,
   PanelMain,
   PanelMainBody,
-  TextContent,
-  Text,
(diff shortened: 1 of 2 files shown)


ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/[groupId]/reset-offset/ResetOffset.tsx
```diff
@@ -215,6 +212,18 @@ export function ResetOffset({
                       type="number"
                       min={0}
                     />
+                    {error && (
```
Comment (MikeEdgar) [inline ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/[groupId]/reset-offset/ResetOffset.tsx:215]: My only concern here is that `error` could be other things besides the custom offset being out of range. The other options are reporting an error from Kafka or that a requested partition is not valid for the topic. I think those other options are much less likely than an out-of-range error, but just to mention that we would be masking them as an out-of-range error.

MikeEdgar (approved):
In the absence of codes identifying the errors more distinctly, this seems reasonable for now. We just need to be careful about changing messages. 

It would probably be good to have a playwright script that triggers the ones that can be triggered via the UI to verify we don't break things inadvertently in the future. That doesn't need to be in this PR right away, but maybe open an issue for it so we remember.

PR #1427, opened 2025-01-31
OpenAPI bug fixes, remove unused properties, ignore split packages

MikeEdgar:
* Fix `@APIResponse` to remove unnecessary `Void` schema (cleans up warning in logs)
* Correct OAS filter to obtain the OIDC URL from configuration rather than the previously-used Quarkus property
* Remove unused Quarkus properties now that the extensions for Kafka and Apicurio are not used (we use the libraries directly rather than via extensions)
* Index the Kafka jar via properties
* Ignore split packages found in several Apicurio jars


MikeEdgar:
IMO we can disregard the lower coverage because it is limited to exceptions.

api/src/main/java/com/github/streamshub/console/api/support/OASModelFilter.java
```diff
@@ -117,12 +119,31 @@ public void filterOpenAPI(OpenAPI openAPI) {
...
-                    .map(schemes -> schemes.get("ConsoleSecurity"))
-                    .ifPresent(scheme -> scheme.setOpenIdConnectUrl(url)));
+        try {
+            Instance<ConsoleConfig> configInstance = CDI.current().select(ConsoleConfig.class);
+
+            if (configInstance.isResolvable()) {
+                consoleConfig = Optional.of(configInstance.get());
+            }
+        } catch (Exception e) {
+            // Ignore, expected at build time where CDI may not be available
```
jankalinic, line 131: Should we at least log it?

MikeEdgar: I looked a bit more closely and it turns out the filter will _not_ run at build time anyway, so I will refactor this slightly. The exception handling doesn't need to be split because of that.

jankalinic (approved):

PR #1429, opened 2025-01-31
fix: locate dependent resources by name

MikeEdgar:
The operator currently uses labels to filter dependent resources when there are multiple instances of a particular type (e.g. Prometheus deployment and Console deployment). This is a problem because the labels may be removed or changed by a user, or could change between releases of this project. 

If the operator cannot locate the resource using the expected labels, it will attempt to create it again, since it believes it does not yet exist. This PR fixes this case by having the operator identify the resources by name rather than by labels.


jankalinic (approved):
Looks great! I think this approach is very good. With the use of `managedby` it's easier to understand when looking at the deployed resources. Thanks.

PR #1463, opened 2025-02-18
fix: avoid use of NEXT_PUBLIC environment variables

MikeEdgar:
`NEXT_PUBLIC` variables are set at build time only and lead to unexpected behavior when they cannot be overridden at run time.


MikeEdgar:
This may need some additional adjustment to ensure no secret information is leaked to the client side - e.g. OIDC client secret.

PR #1475, opened 2025-02-21
Add example for self-signed OCP and fix OICD examples

Frawless:
This PR contains the following changes:
- adding example for Console on OpenShift with internal monitoring and self-signed CA cert
- fix OICD examples to follow CRD schema 
    - `privileges` now requires one from ALL, CREATE, DELETE, GET, LIST, UPDATE
    - `rules` are defined as list of items that contains `resources`, `resourceNames`, `privileges` keys
    - `resourceNames` cannot be empty for Admins as it results into empty cluster list (maybe a bug @MikeEdgar ?)


examples/console/console-security-oidc.yaml
```diff
@@ -107,11 +109,11 @@ spec:
                   - consumerGroups
                   - rebalances
                   - nodes/configs
-              - privileges:
-                  - get
-                  - list
+                privileges:
+                  - GET
+                  - LIST
```
Frawless, line 114: It seems there is also a bug when `resourceNames` is not set. I opened https://github.com/streamshub/console/issues/1476

MikeEdgar (approved):
Thanks @Frawless for the fixes. The privilege parser should be lenient (e.g. allowing both `get` and `GET`), but using the values as specified in the CRDs enum is better :+1:

Frawless:
> Thanks @Frawless for the fixes. The privilege parser should be lenient (e.g. allowing both `get` and `GET`), but using the values as specified in the CRDs enum is better 👍

Unfortunately it didn't work with just lowecase so maybe some wrong config on CRD level? Because `kubectl` didn't allow it based on scheme I think

PR #1487, opened 2025-02-25
Remove cluster name from recently viewed topic card

hemahg:
Remove cluster name from recently viewed topic card
Add topic status to the recently viewed topic card

Screenshot 2025-02-25 at 8 14 15 PM


ui/api/topics/actions.ts
```diff
@@ -177,6 +186,7 @@ export async function setTopicAsViewed(kafkaId: string, topicId: string) {
       topicId,
       // name is included in the `fields[topics]` param list so we are sure it is present
       topicName: topic.attributes.name!,
+      topicStatus: topic.attributes.status!,
```
MikeEdgar, line 189: Just something to be aware of is that the topic status in the overview card will always have the status that it had at the point it was last viewed. Are we OK with that? Having the status be up-to-date would require `getViewedTopics` to be more complex to actually load the topics fresh.

MikeEdgar: Thinking about this again, maybe it would be better to just list the topic name and not the status to avoid it being stale? What do you think @hemahg @shirimordechay ?

hemahg: @MikeEdgar As discussed, I have removed the status field from the table for now.

ui/components/ClusterOverview/ClusterMessages.tsx
MikeEdgar, line 1: I didn't see this referenced in the code, is it needed anymore?

ui/components/ClusterOverview/components/TopicsTable.tsx
```diff
@@ -5,8 +5,58 @@ import { Link } from "@/i18n/routing";
...
+import {
+  CheckCircleIcon,
+  ExclamationCircleIcon,
+  ExclamationTriangleIcon,
+} from "@/libs/patternfly/react-icons";
+import { TopicStatus } from "@/api/topics/schema";
 
-export const TopicsTableColumns = ["name", "cluster"] as const;
+export const TopicsTableColumns = ["name"] as const;
+const StatusLabel: Record<TopicStatus, ReactNode> = {
```
MikeEdgar, line 18: Can these labels be removed too?

MikeEdgar (commented):
Looks good, thanks @hemahg . Just a couple questions.

ui/components/ClusterOverview/components/TopicsTable.tsx
hemahg: I didn't notice it. Thanks for pointing it out! I have removed it.

ui/components/ClusterOverview/ClusterMessages.tsx
hemahg: removed it

MikeEdgar (approved):
Thanks! :+1:

PR #1493, opened 2025-02-27
Improve KRraft support, use node pools and pods for node information

MikeEdgar:
API changes:
- Move nodes collection to its own top-level endpoint under the Kafka cluster
  - Add the usual sorting and filtering support for `/api/kafka/{kafkaId}/nodes`
  - Include value metrics as attributes on each node record. These are the values for status, replica counts, and storage utilization.
- Kafka node status calculations
  - Brokers use state metric, if available. Otherwise, determine a broker status from the node's K8s pod, if possible. Otherwise, default to "Running" when the broker is listed by the Admin describeCluster and a hostname is available.
  - Controller status is simply leader, follower, or lagged follower. A lagged follower is (for now) a follower with a logEndOffset less than the leader's logEndOffset. This will be enhanced in a follow-up to include the last fetch time of the follower together with the node's configured max fetch timeout, or a default if not available.
- Change existing `nodes` attribute on the Kafka cluster to be a relationship with metadata including the total number of brokers and the number of "online" brokers (used in overview screen).

UI changes:
- update Kafka nodes page to include controllers and display roles, status per role, and node pool information.

image


MikeEdgar:
@hemahg , the playwright tests are failing because the "Total Replicas" header is truncated when the table width is narrowed. It appears there is plenty of space remaining based on the status column, but I can't figure out how to make the columns more flexible. Do you have any ideas?

image

hemahg (approved):
LGTM, Thanks for the great work.
I will start with the UI work

PR #1527, opened 2025-03-13
Ensure cluster errors and warnings are displayed even when metrics are unavailable

hemahg:
fix: Display cluster errors and warnings regardless of metrics availability

Previously, errors and warnings were not shown if metrics were unavailable, which was incorrect.  
This update ensures that errors and warnings are always displayed, irrespective of the presence of metrics.


hemahg:
Hi @MikeEdgar, I noticed that the 'Reconciliation Paused' warning was not displayed when metrics were unavailable. Since the backend provides this warning regardless of the metrics configuration, I have updated the code to ensure warnings and errors are shown irrespective of the metrics configuration. Could you please confirm if this is the right approach?

MikeEdgar (approved):
Good catch, thanks @hemahg . Yes, this looks like a good approach.

PR #1536, opened 2025-03-14
[ST] New Kafka examples, install files - STs prequisite

jankalinic:
This PR adds new Kafka example file KafkaTopic and splits KafkaNodePools into supported configuration of seperate broker and controller roles. PR also specifies that Console can be installed without operator and adds folder for OLM operator installation. 

In following PRs there will be also a YAML type operator installation added.


examples/kafka/020-KafkaNodePool-controller-console-nodepool.yaml
```diff
@@ -0,0 +1,18 @@
+---
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaNodePool
+metadata:
+  name: console-nodepool
+  labels:
+    strimzi.io/cluster: console-kafka
```
MikeEdgar, line 7: Since we have two node pools to be used together, let's call this one `console-brokers` and the other one `console-controllers` (or similar). They will need to have distinct names anyway.

MikeEdgar (commented):
It would probably be a good idea to add `systemtests` to the GH workflows that have the `on.pull_request.branches` array to get the CI running while we are using that branch. Wdyt?

jankalinic:
> It would probably be a good idea to add `systemtests` to the GH workflows that have the `on.pull_request.branches` array to get the CI running while we are using that branch. Wdyt?

Sure, shall we run the tests using mvn on k8s cluster? Or do you have any other check/workflow in mind?

MikeEdgar:
> Sure, shall we run the tests using mvn on k8s cluster? Or do you have any other check/workflow in mind?

Nothing in particular just yet. I'm currently only thinking about the existing tests running with the renamed files in the PR, for example. Going ahead with new tests might need something else, if not mvn.

MikeEdgar (approved):
LGTM :+1:

jankalinic:
Thank you for the review, i'll continue to merge and add more PRs.

PR #1545, opened 2025-03-17
[ST] Add ST module structure, basic logger and junit settings, Environment.

jankalinic:
This PR adds gitignore with config.yaml that is used as easy to config environment for systemtesting. PR also adds testing frame resource manager and namespace manager that is yet to be completed. Resource manager should take care of resource creation and deletion with basic waiting utils. Namespace manager should hold namespaces and help with log colletion. Log collection is also a topic for this PR, PR utilizes testframe log collector, which should be enough for testing and failure debug of console.


systemtests/src/main/java/com/github/streamshub/systemtests/resources/NamespaceManager.java
```diff
@@ -0,0 +1,151 @@
...
+import org.apache.logging.log4j.Logger;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class NamespaceManager {
```
Frawless, line 16: Could you please elaborate a little bit more why is this needed for Console tests?

jankalinic: The NamespaceManager ensures proper isolation and cleanup of namespaces during Console tests. It prevents namespace leaks, maintains test suite consistency, and allows efficient tracking of created namespaces and logcollection. Without it, tests could interfere with each other, leading to flaky results.

jankalinic: Is TF collecting logs automatically or do i need to collect them like this? If so, then how do i get the access to namespaces from the TF? From my understanding of TF `KubeResourceManager.STORED_RESOURCES` is the handle you talk about, right?

`
List<String> namespaces = NamespaceManager.getInstance().getListOfNamespacesForTestClassAndTestCase(testClass, testCase);
        if (namespaces.isEmpty()) {
            LOGGER.warn("This test did not have any namespaces");
        } else {
            testCaseCollector.collectFromNamespaces(namespaces.toArray(new String[0]));
        }
`

jankalinic: Also am i missing some info, or is TF capable of preparing  the namespace with pull secrets? if not, what would be the correct approach? I think i would still require to handle these situations, or not?

Frawless: Namespace manager was originally used in Strimzi to handle properly parallel test execution, but even this can be now handled by TF. In case you will find some place for improvement related to this, we can discuss it and implement if needed. but so far it should be in place for most of the usecases.

jankalinic: Cool, I will use that this was then, thanks.

systemtests/pom.xml
```diff
@@ -0,0 +1,226 @@
...
+                                <!-- Exclude LogCollector and Logger from code coverage -->
+                                <exclude>com/github/streamshub/systemtests/logs/*</exclude>
+                                <!-- Exclude unit test utils -->
+                                <exclude>com/github/streamshub/systemtests/unit/*</exclude>
+                            </excludes>
+                            <includeCurrentProject>true</includeCurrentProject>
+                            <outputDirectory>${project.reporting.outputDirectory}/jacoco</outputDirectory>
+                            <formats>
+                                <format>HTML</format>
+                            </formats>
```
MikeEdgar, line 213: Sonar needs the XML report format.
```suggestion
                            <formats>
                                <format>HTML</format>
                                <format>XML</format>
                            </formats>
```

systemtests/pom.xml
```diff
@@ -0,0 +1,226 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>com.github.streamshub</groupId>
+        <artifactId>console</artifactId>
+        <version>0.7.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>systemtests</artifactId>
```
MikeEdgar, line 12: How about using the `console-` prefix? This will match the other modules too. The directory and module names can stay plain `systemtests`.

```suggestion
    <artifactId>console-systemtests</artifactId>
```

systemtests/pom.xml
```diff
@@ -0,0 +1,226 @@
...
+    <description>
+        The intention of this module is to test StreamsHub Console UI with kafka instance deployed and managed by Strimzi Cluster Operator, to simulate a real user scenarios for end-to-end testing.
+    </description>
+
+    <properties>
+        <maven.compiler.source>17</maven.compiler.source>
+        <maven.compiler.target>17</maven.compiler.target>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
MikeEdgar, line 20: These can probably be removed. The parent uses `maven.compiler.release`, which is a newer/preferred way to set the Java version and the source encoding has the same value in the parent.

systemtests/src/main/java/com/github/streamshub/systemtests/cluster/KubeCluster.java
```diff
@@ -0,0 +1,58 @@
+package com.github.streamshub.systemtests.cluster;
+
+import com.github.streamshub.systemtests.logs.LogWrapper;
+import io.skodjob.testframe.clients.KubeClusterException;
+import org.apache.logging.log4j.Logger;
+
+import java.util.Arrays;
+import java.util.Locale;
+
+public interface KubeCluster {
```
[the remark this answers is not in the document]
Frawless: I am asking, because you are adding quite a complex part into the code and I think it can be done in more elegant way.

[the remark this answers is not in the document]
jankalinic: Hmm, so you suggest let's de-class-ify it to reduce the complexity? Having TEST_CLUSTER env variable and then clusterUtils for the mid-testing checks? Although i do not see this part as that much complex.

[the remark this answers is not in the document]
jankalinic: As i mentioned in the previous comment, i like the usability of checking the unstable clusters before each test, so i would like to keep something like that in the STs.

Based on the cluster type i do not want to force user into an installation type (olm or yaml), rather use the info about current cluster to modify the kafka parameters for example. In console examples you can find Kafka yaml example file, which uses `type: ${LISTENER_TYPE}`. This has to be set by user, while for STs it could be determined by the cluster you are using.

While you are logged into a cluster of a certain type, the code i suggest in this PR will ensure that the current context will be used for testing. I liked the idea, with that i'm not saying that there isn't a better way (there is always a better way). 

As you say, using openshift annotation is a good way -> to let's say force some tests to be run only on openshift, or specific installation to be forced, but in this case the installation could be similar, with just a few fine tuning modifications to it by checking the cluster type. So that's why i think having these cluster types implementing kubecluster could be a good way.

At this moment i have no clear path for kind planned, but it could surely follow the same structure.

I think i can rename the variables, you are right that the clusterName might not represent it just like clusterType, i can rename that.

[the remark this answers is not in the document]
Frawless: Discussed offline - @jankalinic will try simple approach as we didn't find reasonable use-case for it now. I will review it after updates.

systemtests/pom.xml
```diff
@@ -0,0 +1,236 @@
...
+                                <exclude>com/github/streamshub/systemtests/constants/*</exclude>
+                                <exclude>com/github/streamshub/systemtests/enums/*</exclude>
+                                <!-- Exclude Environment. It's tested by TF -->
+                                <exclude>com/github/streamshub/systemtests/Environment.class</exclude>
+                                <!-- Exclude LogCollector and Logger from code coverage -->
+                                <exclude>com/github/streamshub/systemtests/logs/*</exclude>
+                                <!-- Exclude unit test utils -->
+                                <exclude>com/github/streamshub/systemtests/unit/*</exclude>
+                                <!-- Exclude exceptions -->
+                                <exclude>com/github/streamshub/systemtests/exceptions/*</exclude>
```
MikeEdgar, line 215: Move these several `exclude` options into `coverage/pom.xml` with the others. The rest of the `report-aggregate` execution can simply be deleted, all reporting is done by the `coverage` module. Keep the `prepare-agent` execution here though.

MikeEdgar (changes requested):
See comments for how to resolve the 0% coverage report.

systemtests/pom.xml
jankalinic: Awesome, will do, thank you.

MikeEdgar (approved):
Sonar is not cooperating with the exclusion criteria, although they are properly being excluded from the Jacoco report. We can go ahead with this one with the understanding that the coverage is going to increase as we add tests to the module.

jankalinic:
I think i adressed all the concerns, leaving it for tomorrow's review. Prepared to merge and continue with follow-up STs.

jankalinic:
Seems like there is no new discussion about the code, so i'm merging this PR for now. Thanks for the review all!

PR #1552, opened 2025-03-18
feat: UI changes for KRaft

hemahg:
feat: UI changes for KRaft

### DIFF
--- a/ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/DistributionChart.tsx
+++ b/ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/DistributionChart.tsx
@@ -138,11 +138,9 @@ export function DistributionChart({
                   </Tooltip>
                 </DescriptionListTerm>
                 <DescriptionListDescription>
-                  <Link href={`nodes/${leadController}`}>
-                    {t("DistributionChart.lead_controller_value", {
-                      leadController,
-                    })}
-                  </Link>
+                  {t("DistributionChart.lead_controller_value", {
+                    leadController,
+                  })}
                 </DescriptionListDescription>
               </DescriptionListGroup>
             </DescriptionList>
--- a/ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/DistributionChart.tsx
+++ b/ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/DistributionChart.tsx
@@ -13,22 +13,43 @@ import {
   CardBody,
   CardHeader,
   CardTitle,
+  DescriptionList,
+  DescriptionListDescription,
+  DescriptionListGroup,
+  DescriptionListTerm,
+  Grid,
+  GridItem,
+  Icon,
   ToggleGroup,
   ToggleGroupItem,
   Tooltip,
 } from "@/libs/patternfly/react-core";
-import { HelpIcon } from "@/libs/patternfly/react-icons";
+import {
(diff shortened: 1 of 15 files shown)


ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/page.tsx
```diff
@@ -27,56 +18,180 @@ export async function generateMetadata() {
...
+    }
+
+    if (node.attributes.roles.includes("controller")) {
+      nodeCounts.totalControllers++;
+    }
+
+    if (node.attributes.metadataState?.status === "leader") {
+      nodeCounts.leadControllerId = node.id;
+    }
+  });
```
Comment (MikeEdgar) [inline ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/page.tsx:169]: This will only tally the nodes on the current page. The `meta.summary.statuses` will give the counts for all nodes regardless of filtering or current page. I also realize now that we need the lead controller's ID in the meta section also so that it's usable even when the leader is not on the current page. I'll open a PR with that additional meta property.

```javascript
  nodeCounts.totalNodes = Object.
    values(nodes.meta.summary.statuses.combined ?? 0).
    reduce((tally, value) => tally + value, 0);
  nodeCounts.totalBrokers = Object.
    values(nodes.meta.summary.statuses.brokers ?? 0).
    reduce((tally, value) => tally + value, 0);
  nodeCounts.totalControllers = Object.
    values(nodes.meta.summary.statuses.controllers ?? 0).
    reduce((tally, value) => tally + value, 0);
```

MikeEdgar (commented):
It looks like the node count is appearing in the role column:
image

ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/DistributionChart.tsx
```diff
@@ -68,125 +90,205 @@ export function DistributionChart({
...
+                  >
+                    <HelpIcon />
+                  </Tooltip>
+                </DescriptionListTerm>
+                <DescriptionListDescription>
+                  <Link href={`nodes/${leadController}`}>
+                    {t("DistributionChart.lead_controller_value", {
+                      leadController,
+                    })}
+                  </Link>
```
Comment (MikeEdgar) [inline ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/DistributionChart.tsx:145]: Controller-only node configuration not retrievable - suggest to just drop the link.

```suggestion
                  {t("DistributionChart.lead_controller_value", {
                    leadController,
                  })}
```

ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/NodesTable.tsx
```diff
@@ -1,259 +1,419 @@
...
+              case "nodePool":
+                return <Th key={key}>{t("nodes.nodePool")}</Th>;
+            }
+          }}
+          renderCell={({ column, key, row, Td }) => {
+            switch (column) {
+              case "id":
+                return (
+                  <Td key={key} dataLabel={"Node ID"}>
+                    <Link href={`nodes/${row.id}`}>{row.id}</Link>
```
Comment (MikeEdgar) [inline ui/app/[locale]/(authorized)/kafka/[kafkaId]/nodes/NodesTable.tsx:179]: Controller-only node configuration not retrievable - make link available only when the broker role is present. Maybe we need a tool-tip to indicate the controller configuration is not available?

```suggestion
                    {row.attributes.broker
                        ? <Link href={`nodes/${row.id}`}>{row.id}</Link>
                        : <>{row.id}</>
                    }
```

MikeEdgar (approved):
Looks good, thanks @hemahg

PR #1554, opened 2025-03-20
Enable configuration of trust stores via the YAML file

### LINKED ISSUE #1335: Allow configuration of truststores via configuration YAML
The truststore support for Apicurio, Prometheus, and (WIP) OIDC all depend on the use of Quarkus environment variables. It would be better to allow truststores to be configured in the application's configuration YAML and the API server creates and registers Quarkus `TlsConfiguration` at start-up time. We can use the built-in `VertxCertificateHolder` implementation for this.

MikeEdgar:
Closes #1335 

- Add support for trust stores to be configured in the API's configuration YAML for metrics sources, schema registry, and the OIDC provider. This eliminates the need for using environment variables or command line parameters to configure the Quarkus TLS registry for these connections, making the TLS configurations entirely private to the API module based on the configuration YAML
- Update the operator and UI to use the new configurations and remove the previous mappings to volumes/mounts/environment variables.
- Updates to tests as needed to use the new configuration structure


api/src/main/java/com/github/streamshub/console/api/support/factories/ConsoleConfigFactory.java
```diff
@@ -143,4 +148,6 @@ private String resolveValue(String value) {
 
         return value;
     }
+
```
jankalinic, line 151: shall we setup a checkstyle rule for more than one empty lines? i had issue with that in my PR as well

jankalinic (approved):
Looks good, did only static checks, did not run the code. I need to write STs for this. (already in my todo). Thanks for the PR!

PR #1555, opened 2025-03-20
Add description text to checkbox selection option

hemahg:
Add description text to checkbox selection option as the kRaft changes require description text in the checkbox selection option


MikeEdgar (approved):
LGTM

PR #1558, opened 2025-03-20
fix: use common aggregate coverage data file in root, remove editable

### LINKED ISSUE #1557: Disable "editable" generated classes
There are several remaining classes in the common configuration package and in the operator CRD packages that use `@io.sundr.builder.annotations.Buildable` without setting `editableEnabled` to false. This results in unused classes being generated and we should have this disabled throughout the project.

MikeEdgar:
Closes #1557 

- Remove editable on `@Buildable` classes that do not yet disable it
- Aggregate coverage data to common file in root project. The operator module will run the report (via Quarkus Jacoco support) which generates the XML data submitted to Sonar
- Fix flaky test race condition in `NodesResourceIT`


jankalinic (approved):
Looks good, thanks.

PR #1578, opened 2025-03-26
fix: resolve sonar issues, correct unsorted configs response

MikeEdgar:
Resolve a number of issues/code smells identified by Sonar. Also includes a fix to the config retrieval service to fix the sorting of properties by key.


MikeEdgar:
I'm ignoring the < 80% coverage on this one because the uncovered lines are all exception/edge cases that are not worth the trouble mocking in a test to cover.

PR #1584, opened 2025-03-27
Fix consumer group story

hemahg:
Pagination was missing in Storybook because the total count was passed as a parameter. This PR fixes the issue, ensuring that the story displays correctly and that the consumer group works as expected.


hemahg:
@MikeEdgar, hope this PR makes the pagination visible in storybook

MikeEdgar (changes requested):
This looks good. I left a few suggestions. Using `(i * 1000)` for the lag makes the result stable and Chromatic won't identify each build as changed from the baseline - similar to how we use hard-coded dates in other stories.

ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/ConsumerGroupsTable.stories.tsx
```diff
@@ -12,1753 +12,67 @@ export default {
...
       },
     ],
   },
+});
+
+export const ConsumerGroups: Story = {
+  args: {
+    perPage: 20,
+    consumerGroups: (() => {
+      const groups = Array.from({ length: 30 }, (_, i) =>
```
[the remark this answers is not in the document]
Comment (hemahg) [reply ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/ConsumerGroupsTable.stories.tsx]: I have kept length to be 21 just to show pagination
 const groups = Array.from({ length: 21 }, (_, i) =>

MikeEdgar (approved):

hemahg:
That was lightning fast!

ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/ConsumerGroupsTable.stories.tsx
```diff
@@ -12,1753 +12,67 @@ export default {
...
+        generateConsumerGroup(
+          `console-datagen-group-${i}`,
+          ["STABLE", "EMPTY"][i % 2],
+          i * 1000,
+          i * 1000,
+          i % 3 === 0 ? i * 1000 : undefined,
+        ),
+      );
+      return groups.length;
+    })(),
```
Comment (MikeEdgar) [inline ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/ConsumerGroupsTable.stories.tsx:76]: One question actually on this part - is there any different between the lambda and simply `total: 21` ?

Comment (hemahg) [reply ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/ConsumerGroupsTable.stories.tsx:76]: Yes, we can simply write the total number instead of using a lambda function, as the array creation is unnecessary. I just did it to match the number of consumers.

Comment (MikeEdgar) [reply ui/app/[locale]/(authorized)/kafka/[kafkaId]/consumer-groups/ConsumerGroupsTable.stories.tsx]: No worries, just confirming my understanding of it. I guess the number makes it most obvious, but I'm fine with either.

PR #1585, opened 2025-03-28
[ST] Add strimzi setup class, reverse coverage logic, add test-frame snapshot

jankalinic:
- This PR adds basic strimzi cluster operator setup. 
- Also it would be overwhelming to exclude, so this PR includes a reverse logic -> including UTs instead of excluding files.
- PR adds test-frame snapshot repo which helps cover newly found and fixed items. + PR updates depricated methods to a newly enforced ones.
- PR switches pioneer for a powermock for a smoother system environment variables mocking without unchecked operations.
- PR switches environment variable nulls for empty strings, which makes the string checks shorter and if statements more readable
- PR adds Constants, labels and resourcekinds


systemtests/src/main/java/com/github/streamshub/systemtests/setup/StrimziOperatorSetup.java
```diff
@@ -0,0 +1,306 @@
...
+
+    private OperatorGroup getStrimziOperatorGroup() {
+        return new OperatorGroupBuilder()
+            .editOrNewMetadata()
+                .withName(Constants.CONSOLE_OPERATOR_GROUP_NAME)
+                .withNamespace(this.deploymentNamespace)
+                .withLabels(Collections.singletonMap(Labels.APP, Constants.STRIMZI_NAME))
+            .endMetadata()
+            .build();
+    }
```
MikeEdgar, line 154: Is the `OperatorGroup` necessary when installing without OLM?

systemtests/src/main/java/com/github/streamshub/systemtests/constants/ResourceKinds.java
```diff
@@ -6,37 +6,41 @@ private ResourceKinds() {}
...
     // Kafka
     // ----------
-    public static final String KAFKA_TOPIC = "kafkatopic";
-    public static final String KAFKA = "kafka";
-    public static final String KAFKA_USER = "kafkauser";
-    public static final String KAFKA_NODE_POOL = "kafkanodepool";
+    public static final String KAFKA_TOPIC = "KafkaTopic";
+    public static final String KAFKA = "Kafka";
+    public static final String KAFKA_USER = "KafkaUser";
+    public static final String KAFKA_NODE_POOL = "KafkaNodePool";
```
im-konge, line 39: You can use the RESOURCE_KIND from each of the resource from the Strimzi's API module.

systemtests/src/main/java/com/github/streamshub/systemtests/constants/ResourceKinds.java
```diff
@@ -6,37 +6,41 @@ private ResourceKinds() {}
...
+
     // ----------
     // Console
     // ----------
-    public static final String CONSOLE = "console";
+    public static final String CONSOLE = "Console";
```
im-konge, line 44: Do you have kind in the Console CR?

systemtests/src/main/java/com/github/streamshub/systemtests/setup/StrimziOperatorSetup.java
```diff
@@ -0,0 +1,306 @@
...
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class StrimziOperatorSetup {
+    private static final Logger LOGGER = LogWrapper.getLogger(StrimziOperatorSetup.class);
+    private static final KubeResourceManager RESOURCE_MANAGER = KubeResourceManager.get();
```
im-konge, line 45: out of curiosity, why do you have it as static here?

systemtests/src/main/java/com/github/streamshub/systemtests/setup/StrimziOperatorSetup.java
```diff
@@ -0,0 +1,306 @@
...
+                Files.write(tempFile, SetupUtils.getYamlFileContent(fileUrl).getBytes());
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+            yamlFiles.add(tempFile.toFile());
+        }
+        return yamlFiles;
+    }
+
+    private List<String> extractYamlFileLinksFromGithubDir() {
```
im-konge, line 261: Maybe it's something common that you can use for other operators/components/files as well? IDK if you have something common. There are some things in test-frame as well -> e.g. `TestFrameUtils#getFileFromResourceAsStream` etc.
Maybe it will be useful for you as well?

systemtests/src/test/java/com/github/streamshub/systemtests/system/AbstractST.java
```diff
@@ -72,12 +81,28 @@ public abstract class AbstractST {
     }
 
     @BeforeAll
-    void beforeAll() {
-        LOGGER.info("=========== AbstractST - BeforeAll ===========");
+    void setupTestSuite() {
+        LOGGER.info("=========== AbstractST - BeforeAll - Setup TestSuite ===========");
```
im-konge, line 85: Maybe `TestVisualSeparatorExtension` can be used from test-frame? Depends if you want it that way. Just wanted to mention it, so you are aware of it.

systemtests/src/main/java/com/github/streamshub/systemtests/setup/StrimziOperatorSetup.java
```diff
@@ -0,0 +1,306 @@
...
+        RESOURCE_MANAGER.deleteResource(getBundleCrds());
+        RESOURCE_MANAGER.deleteResource(getBundleClusterRoles());
+        RESOURCE_MANAGER.deleteResource(getBundleServiceAccount());
+        RESOURCE_MANAGER.deleteResource(getBundleRoleBindings());
+        RESOURCE_MANAGER.deleteResource(getBundleClusterRoleBindings());
+        RESOURCE_MANAGER.deleteResource(getBundleConfigMap());
+        RESOURCE_MANAGER.deleteResource(getBundleDeployment());
+    }
+
+    private ClusterRoleBinding[] getStrimziWatchAllCrbs() {
```
MikeEdgar, line 96: Some of these bindings are in the Strimzi resources aren't they? Maybe there are differences that I'm not spotting though.

systemtests/src/main/java/com/github/streamshub/systemtests/utils/SetupUtils.java
```diff
@@ -0,0 +1,84 @@
...
+            return mapper.readValue(yamlFile, c);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    // --------------
+    // Namespace
+    // --------------
+    public static void copyImagePullSecrets(String targetNamespace) {
```
MikeEdgar, line 55: Will there be any private images included in the testing?

systemtests/src/main/java/com/github/streamshub/systemtests/constants/ResourceKinds.java
```diff
@@ -6,37 +6,41 @@ private ResourceKinds() {}
...
-    public static final String CATALOG_SOURCE = "catalogsource";
-    public static final String SUBSCRIPTION = "subscription";
-    public static final String INSTALL_PLAN = "installplan";
-    public static final String CLUSTER_SERVICE_VERSION = "clusterserviceversion";
-    public static final String OPERATOR_GROUP = "operatorgroup";
-    public static final String OPERATOR = "operator";
+    public static final String CUSTOM_RESOURCE_DEFINITION = "CustomResourceDefinition";
+    public static final String CONFIG_MAP = "ConfigMap";
+    public static final String DEPLOYMENT = "Deployment";
+    public static final String DEPLOYMENT_CONFIG = "DeploymentConfig";
```
MikeEdgar, line 12: The `DeploymentConfig` might never be used.

MikeEdgar (commented):
First pass, left a few comments/questions.

systemtests/src/main/java/com/github/streamshub/systemtests/setup/StrimziOperatorSetup.java
jankalinic: For a more simple call usage for the whole class.

systemtests/src/main/java/com/github/streamshub/systemtests/constants/ResourceKinds.java
jankalinic: Not that i am aware in ConsoleResource, no.

systemtests/src/main/java/com/github/streamshub/systemtests/setup/StrimziOperatorSetup.java
jankalinic: I guess i could use configFromYaml then, thanks for bringing that up.

systemtests/src/main/java/com/github/streamshub/systemtests/utils/SetupUtils.java
jankalinic: There is currently a public image in the config.yaml file. This only allowed usage of custom test client with specific kafka version. But i guess for our usage, we could go only the public image, so we can get rid of all of this around test client image being a custom one.

systemtests/src/main/java/com/github/streamshub/systemtests/setup/StrimziOperatorSetup.java
im-konge: I personally have it in the variable in `AbstractST` so in the STs I'm just doing `kubeResourceManager.create...`.
But in all other classes that are not in the `test` folder I'm using the `KubeResourceManager.get()`

I was mainly curious why you wanted to do it like that in the `StrimziOperatorSetup`. You want to do this in every class you will create or delete something?

(again, I'm just asking, if you want to do it like this, okay. I'm not sure if it can cause some issues or not.)

jankalinic: Sure, let's keep it united everywhere, it might be better to call it then define it all around.

systemtests/src/test/java/com/github/streamshub/systemtests/system/AbstractST.java
jankalinic: You can see that even when separator logs the method name, it is more visible when the line you marked is logged as well. Or at least that is my opinion.

im-konge: I guess it's up to you then :)

coverage/pom.xml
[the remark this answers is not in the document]
MikeEdgar: It will probably be rare for PRs to change both the application code and the STs, so being aware that ST changes will result in less than 80% coverage is fine with me. We can exclude everything in the ST package if you like, and gradually add things back as tests are running. For the STs, coverage metrics are basically only just helper since we have less concern about untested application functionality.

systemtests/src/main/java/com/github/streamshub/systemtests/setup/StrimziOperatorSetup.java
jankalinic: Had to check, but it seems like i should remove it from here. Thanks!

systemtests/src/main/java/com/github/streamshub/systemtests/setup/StrimziOperatorSetup.java
jankalinic: I think i had to resort to this because of strimzi not watching all namespaces, not by default only using STRIMZI_NAMESPACE:"*" i think. Follows ->
https://strimzi.io/docs/operators/latest/deploying#deploying-cluster-operator-to-watch-whole-cluster-str

systemtests/src/main/java/com/github/streamshub/systemtests/constants/ResourceKinds.java
im-konge: @MikeEdgar would it make sense to add the kind here: https://github.com/streamshub/console/blob/main/operator/src/main/java/com/github/streamshub/console/api/v1alpha1/Console.java
?

I mean, the `kind` in the CR is taken from the class name? Usually I saw that these CR classes has the `KIND` or `RESOURCE_KIND` constant just to provide it to users that are also using the API. Or if it should be taken just from the display name of the class (IDK how it is handled in the generation)?

systemtests/src/test/java/com/github/streamshub/systemtests/system/AbstractST.java
jankalinic: Right, i'd keep this one in.

systemtests/src/main/java/com/github/streamshub/systemtests/constants/ResourceKinds.java
MikeEdgar: Yeah it's just the class name - `Class#getSimpleName` would provide it.

MikeEdgar: Fabric8 will also get it for you if you use `HasMetadata.getKind(Console.class)`.

systemtests/pom.xml
```diff
@@ -23,15 +23,29 @@
         <playwright.version>1.47.0</playwright.version>
         <!-- Mocking environment -->
         <mockito.version>5.15.2</mockito.version>
-        <junit-pioneer.version>2.3.0</junit-pioneer.version>
+        <powermock.version>2.0.9</powermock.version>
         <!-- Used for test-frame -->
-        <test-frame.version>0.8.0</test-frame.version>
+        <test-frame.version>0.11.0-SNAPSHOT</test-frame.version>
```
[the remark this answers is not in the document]
jankalinic: I think i can leave it for now, but remove once released.

jankalinic:
Hopefully last commit added all the fixes for the reviewed bits. If you have time, could you please take a look?

systemtests/src/main/java/com/github/streamshub/systemtests/constants/Constants.java
```diff
@@ -0,0 +1,71 @@
...
+
+
+     // --------------------------------
+    // ------------ Console -----------
+    // --------------------------------
+    /**
+     * CRDs
+     */
+    public static final String CONSOLE_CRD_API_NAME = "consoles.console.streamshub.github.com";
+    public static final String CONSOLE_CRD_API_VERSION = "console.streamshub.github.com/v1alpha1";
```
MikeEdgar, line 68: Just to mention it, `HasMetadata` also has methods to obtain these if you want. E.g.
```suggestion
    public static final String CONSOLE_CRD_API_NAME = HasMetadata.getFullResourceName(Console.class);
    public static final String CONSOLE_CRD_API_VERSION = HasMetadata.getGroup(Console.class) + "/" + HasMetadata.getVersion(Console.class);
```

MikeEdgar (approved):

systemtests/src/main/java/com/github/streamshub/systemtests/setup/StrimziOperatorSetup.java
```diff
@@ -0,0 +1,291 @@
...
+                    .withName(Constants.STRIMZI_CO_NAME)
+                    .withNamespace(this.deploymentNamespace)
+                    .build())
+                .build()
+        };
+    }
+
+    private ConfigMap getBundleConfigMap() {
+        return new ConfigMapBuilder(TestFrameUtils.configFromYaml(strimziResources.stream()
+            .filter(file -> file.getName().matches(".*050-ConfigMap-.*\\.yaml")).toList().get(0), ConfigMap.class))
```
MikeEdgar, line 143: I know it seems silly, but it might be good to change the `matches` to `startsWith` and drop the `yaml` part since these regex were all flagged as a security hotspot.
```suggestion
            .filter(file -> file.getName().startsWith("050-ConfigMap-")).toList().get(0), ConfigMap.class))
```

systemtests/src/main/java/com/github/streamshub/systemtests/constants/Constants.java
jankalinic: Okay, i'll change even the other resources, if possible, using HasMetadata. Thanks

jankalinic:
@im-konge Would you have anything against the current state?

im-konge:
Will have a look now. Sorry

im-konge (approved):
LGTM, thanks.

PR #1598, opened 2025-04-02
[ST] Console installation

jankalinic:
PR contains:
 * Chages to config.yaml envs, because `TestEnvironmentVariables` take `""` as an empty value and do not replace it with defaults like null values. So this should save some future issues. Also use HELM to install strimzi, so the env only defines strimzi version now. Set test clients imge to latest stable released
 * Addition of `OLM_CATALOG_NAMESPACE` env for more flexible user-side catalogsource setup options.
 * Addition of  `OperatorGroup` install file into example paths - used for OLM deployment of console operator.
 * Implementation of `ConsoleOperatorSetup` focusing only on the operator installation both YAML and OLM ways  using `InstallConfig` - instance deployment will be included in the next PR
 * Removal of teardown method for `StrimziOperatorSetup` as `KubeResourceManager` can take care of all created resources after the tests in the annotated TestFrame `@AfterAll` which is executed after the `AbstractST`.
 * Addition of `WaitUtils` aimed at waiting for conditions of resources to be fulfilled, mostly their readiness, replicas count etc.
 * Addition of `ResourceUtils` generic listing methods and mapping to provided k8s resource kind type.
 * Removal of yaml reading method in `SetupUtils` and instead created one, that resolves whether `CONSOLE_OPERATOR_BUNDLE_URL` is a local path or an internet URL
 *Addition of setup exceptions to not throw only general RuntimeException


systemtests/src/main/java/com/github/streamshub/systemtests/utils/WaitUtils.java
```diff
@@ -0,0 +1,28 @@
...
+
+public class WaitUtils {
+
+    private WaitUtils() {}
+
+    // ------------
+    // Deployment
+    // ------------
+    public static void waitForDeploymentWithPrefixIsReady(String namespaceName, String deploymentNamePrefix) {
+        Wait.until(String.format("creation of Deployment with prefix: %s in Namespace: %s", deploymentNamePrefix, namespaceName),
```
jankalinic, line 17: Wait until + this string, may seem misleading -> combined with the cause, but for logging in `Wait.until` there is used a String ("Waiting for: " + cause)

MikeEdgar: What is the advantage of `io.skodjob.testframe.wait.Wait` versus Awaitility?

systemtests/src/main/java/com/github/streamshub/systemtests/utils/WaitUtils.java
```diff
@@ -0,0 +1,28 @@
...
+            TestFrameConstants.GLOBAL_POLL_INTERVAL_1_SEC, TestFrameConstants.GLOBAL_TIMEOUT_MEDIUM,
+            () -> {
+                Deployment dep = listKubeResourcesByPrefix(Deployment.class, namespaceName, deploymentNamePrefix).get(0);
+                if (dep == null || dep.getStatus() == null || dep.getStatus().getAvailableReplicas() == null ||
+                    !dep.getStatus().getAvailableReplicas().equals(dep.getSpec().getReplicas())) {
+                    return false;
+                }
+                return true;
```
MikeEdgar, line 25: Take a look at `io.fabric8.kubernetes.client.readiness.Readiness.isDeploymentReady(Deployment)`

systemtests/src/main/java/com/github/streamshub/systemtests/utils/SetupUtils.java
```diff
@@ -27,8 +30,17 @@ public static String getYamlFileContent(String fileUrl) {
                 content.append(inputLine).append("\n");
             }
         } catch (IOException e) {
-            throw new RuntimeException("Cannot download YAML content from url: " + fileUrl, e);
+            throw new FileOperationException("Cannot download YAML content from url: " + fileUrl, e);
         }
         return content.toString();
```
MikeEdgar, line 35: Just a nit, you can probably simplify this method by using `InputStream#readAllBytes`.
For example:
```java
try (var stream = new URL(fileUrl).openStream()) {
    return new String(stream.readAllBytes(), StandardCharsets.UTF_8);
} catch (IOException e) {
    throw new FileOperationException("Cannot download YAML content from url: " + fileUrl, e);
}
```

systemtests/src/main/java/com/github/streamshub/systemtests/utils/WaitUtils.java
jankalinic: Looking at `isDeploymentReady` and noticed i still need to check for deployment not to be null, before passing it to that function. Thanks!

systemtests/src/main/java/com/github/streamshub/systemtests/utils/WaitUtils.java
jankalinic: I think i went through the basic differences and i'd say the way of logging, onTimeout runnables and exceptions description are better with `Wait`. However you know more about the Awaitility then i do. Would you say that it is better? 

For example what about failing with custom messages that will ease the debugging - like "Failed waiting for Pod {} to roll in namespace {}"?

jankalinic: Discussed offline - decided to keep `Wait.until` as it's a slightly more fitting the needs here.

systemtests/src/main/java/com/github/streamshub/systemtests/setup/StrimziOperatorSetup.java
```diff
@@ -233,10 +222,11 @@ private List<File> fetchStrimziYamlFiles() {
             LOGGER.debug("Fetching Strimzi YAML file: {}", fileName);
             Path tempFile = null;
             try {
-                tempFile = Files.createTempFile(fileName + "-tmp_", ".yaml");
-                Files.write(tempFile, SetupUtils.getYamlFileContent(fileUrl).getBytes());
+                FileAttribute<Set<PosixFilePermission>> attr = PosixFilePermissions.asFileAttribute(PosixFilePermissions.fromString("rwx------"));
+                tempFile = Files.createTempFile(fileName + "-tmp_", ".yaml", attr);
+                Files.write(tempFile, SetupUtils.getYamlContentFromUrl(fileUrl).getBytes());
```
im-konge, line 227: I would create method for this, as it seems you are using this on few places.

systemtests/src/main/java/com/github/streamshub/systemtests/utils/ConsoleUtils.java
```diff
@@ -0,0 +1,11 @@
+package com.github.streamshub.systemtests.utils;
+
+import com.github.streamshub.systemtests.Environment;
+
+public class ConsoleUtils {
+    private ConsoleUtils() {}
+
+    public static String getConsoleOperatorName() {
+        return Environment.CONSOLE_DEPLOYMENT_NAME + "-operator";
```
im-konge, line 9: Will it be always like this? I mean, same for OLM and Yaml manifests installation? For the YAML (Bundle) installation this will work, but the OLM deployment name contains version as well or?

im-konge (commented):
Few nits/questions. Thanks.

systemtests/config.yaml
```diff
@@ -28,7 +28,9 @@ CLIENT_TYPE: kubectl
 # OLM
 # -------------------------------------------
 # Name of the CatalogSource for Console Operator (OLM install)
-CONSOLE_CATALOG_SOURCE_NAME: streamshub-console-catalog
+CONSOLE_CATALOG_SOURCE_NAME: strimzi-source
```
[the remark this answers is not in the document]
jankalinic: Good, thanks i must have missed it in the last commits.

systemtests/src/main/java/com/github/streamshub/systemtests/setup/ConsoleOperatorSetup.java
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    // ------
+    // Bundle
+    // ------
+    private List<File> getConsoleBundleResources() {
+        List<File> tempFiles = new ArrayList<>();
+        String[] yamlFiles;
+
+        if (Environment.CONSOLE_OPERATOR_BUNDLE_URL.startsWith("http://") || Environment.CONSOLE_OPERATOR_BUNDLE_URL.startsWith("https://")) {
```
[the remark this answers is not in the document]
jankalinic: Right, i believe at this point, there is no yaml file containing the crds, crbs... and operator deployment for the YAML bundle install, so user can now either download the release then modify the file and reference it locally or use directly the url to the operator.yaml that is in the release. It could also serve for the future upgrade purpose like you said.

systemtests/config.yaml
```diff
@@ -39,24 +41,22 @@ CONSOLE_OLM_DEPLOYMENT_NAME: streamshub-console-operator
...
 # Kafka
 # -------------------------------------------
 # Kafka version used in system tests
 ST_KAFKA_VERSION: 3.9.0
 # Image for test client containers in system tests
-TEST_CLIENTS_IMAGE: quay.io/strimzi-test-clients/test-clients:latest-kafka-3.9.0
-# Secret used for pulling testing clients image
-SYSTEM_TEST_STRIMZI_IMAGE_PULL_SECRET: ""
\ No newline at end of file
+TEST_CLIENTS_IMAGE: quay.io/strimzi-test-clients/test-clients:latest-kafka-3.9.0
```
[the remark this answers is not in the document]
jankalinic: Thanks, i thought these were the latest released. But for now i guess i'll use  `0.10.0-kafka-3.9.0` . Thanks for the tip!

systemtests/src/main/java/com/github/streamshub/systemtests/setup/ConsoleOperatorSetup.java
```diff
@@ -0,0 +1,284 @@
...
+            yamlFiles = SetupUtils.getYamlContentFromUrl(Environment.CONSOLE_OPERATOR_BUNDLE_URL).split("---");
+        } else {
+            yamlFiles = SetupUtils.getYamlContentFromFile(Environment.CONSOLE_OPERATOR_BUNDLE_URL).split("---");
```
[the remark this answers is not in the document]
jankalinic: No use elsewhere, only this instance. I'd keep that here as it relates only to the console setup.

systemtests/src/main/java/com/github/streamshub/systemtests/utils/ConsoleUtils.java
jankalinic: Well, technically yes. For YAML as you mention it is whatever i set. So i took the one used for OLM, this is as you say `prefix` for the operator deployment. In current version it's `streamshub-console-operator-v0.6.4`

im-konge: It depends on what you will use it. If for something like "give me deployment name by this prefix" then it will work all the time. If you will assume that this is the actual full deployment name, you will have issues with that. Just my two cents.

systemtests/src/main/java/com/github/streamshub/systemtests/setup/ConsoleOperatorSetup.java
```diff
@@ -0,0 +1,284 @@
...
+                    tempFile = Files.createTempFile("console-" + fileName + "-tmp_", ".yaml", attr);
+                    Files.write(tempFile, file.trim().getBytes(StandardCharsets.UTF_8));
+                } catch (IOException e) {
+                    throw new FileOperationException("Failed to create temp file: " + fileName, e);
+                }
+
+                tempFiles.add(tempFile.toFile());
+                LOGGER.debug("Created temp file: {}", tempFile.toAbsolutePath());
+            }
+        }
```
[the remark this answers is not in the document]
jankalinic: Hmm, but that is mapping it to the object. I would need to restructure this into -> do not hold any files, use multiobject parse just to get one resource kind each time i want a resource. Or do you have any other design in mind?

systemtests/src/main/java/com/github/streamshub/systemtests/setup/StrimziOperatorSetup.java
jankalinic: Sure, can do

systemtests/src/main/java/com/github/streamshub/systemtests/utils/ConsoleUtils.java
jankalinic: So would you suggest handling all the operations by label? Or some different method?

jankalinic: Okay lets make it more dynamic. In case of OLM install i can get the full name after the deployment process by its prefix.

systemtests/src/main/java/com/github/streamshub/systemtests/setup/ConsoleOperatorSetup.java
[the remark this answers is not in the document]
im-konge: From what I understood from the code, you are anyway copying it to tmp files and then reading the files one by one or? I mean, I think it's possible to just cut the file into pieces and handle each resource one by one.
Then you will not need to filter it as you are doing there, you can just go through the resources and apply them based on the type. Because as you have it now, it goes through the whole list of files many times, which is not much efficient FMPOV.

systemtests/src/main/java/com/github/streamshub/systemtests/utils/ConsoleUtils.java
im-konge: You can also have method that based on installation type, you will either return the constant (for Yaml and Helm possibly) and then for OLM it returns the deployment name based on prefix in Namespace (so you list all Deployments in the particular Namespace and return the matching Deployment).

Or, during the installation you can actually configure some object and then return it from the Setup class or something.

systemtests/src/main/java/com/github/streamshub/systemtests/setup/ConsoleOperatorSetup.java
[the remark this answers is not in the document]
MikeEdgar: The `KubernetesClient` has a `load` class that takes an `Inputstream` and returns a list of `HasMetadata`. Looping over them to find/modify/apply those might be cleaners than the string and temp file handling.

jankalinic:
In last commits, due to change of the getting:

- added and modified installation type, so that there is more flexibility to add new install type if needed
- switched to latest released test clients
- switched to strimzi operator version instead of full url
- modified the fetching mechanism, so that no temporary files are created and resources are rather stored in list of strings, each representing one resource, which are then parsed if the kind is matching
- changed operatorDeploymentName so that it is set to a full name in case of OLM installation after it is deployed
- added small URL validating
- simplified getting separate links from the strimzi github using HTML doc parse
 
+ 
- switched to newly released test-frame version
- fixed logging of env variables after setting log level dynamically

systemtests/src/main/java/com/github/streamshub/systemtests/utils/ConsoleUtils.java
jankalinic: > For OLM it returns the deployment name based on prefix in Namespace (so you list all Deployments in the particular Namespace and return the matching Deployment).

i like this one, hopefully my implementation is what you meant.

systemtests/src/main/java/com/github/streamshub/systemtests/setup/ConsoleOperatorSetup.java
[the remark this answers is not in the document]
jankalinic: Removed creating temp files in the latest commit. Would the latest design be okay with you?

systemtests/src/main/java/com/github/streamshub/systemtests/setup/StrimziOperatorSetup.java
```diff
@@ -226,66 +216,27 @@ private ClusterRoleBinding[] getBundleClusterRoleBindings() {
...
-                while (matcher.find()) {
-                    final String rawUrl = "https://raw.githubusercontent.com" + matcher.group(1)
-                        .replace("/blob/", "/");
-                    if (!yamlFileLinks.contains(rawUrl)) {
-                        yamlFileLinks.add(rawUrl);
-                    }
-                }
+                // Github row locator of current dir
+                List<String> yamlFileLinks = Jsoup.parse(new URL(resourceURL), 10_000)
+                    .select("tr.react-directory-row div.react-directory-filename-cell a.Link--primary")
```
MikeEdgar, line 225: This seems like it could be brittle by relying on the GitHub UI. Isn't this fetching the same files that are concatenated and attached to the Strimzi releases?

jankalinic: It's possible to fetch from the release yaml. My initial idea was to be able to work with all strimzi branches, but they are release branches anyway. So i'll change it to that one specific file. Thanks!

jankalinic:
In last commit, i removed YAML fetching altogether and used kubernetes load to map objects. Also removed unused exceptions and modified config.yaml.

systemtests/src/main/java/com/github/streamshub/systemtests/setup/StrimziOperatorSetup.java
```diff
@@ -177,12 +168,16 @@ private Deployment getBundleDeployment() {
                     .endSpec()
                 .endTemplate()
             .endSpec()
-            .build();
+            .build())
+            .toList()
+            .get(0);
```
MikeEdgar, line 173:
```suggestion
            .findFirst()
            .orElseThrow();
```

systemtests/src/main/java/com/github/streamshub/systemtests/setup/StrimziOperatorSetup.java
```diff
@@ -139,27 +120,37 @@ private ClusterRoleBinding[] getStrimziWatchAllCrbs() {
...
+            .filter(ConfigMap.class::isInstance)
+            .map(ConfigMap.class::cast)
+            .map(r -> new ConfigMapBuilder(r)
+                .editMetadata()
+                    .withNamespace(this.deploymentNamespace)
+                    .withName(this.deploymentName)
+                .endMetadata()
+                .build())
+            .findFirst()
+            .orElseThrow();
```
MikeEdgar, line 133: Not required, but you could cut down some of this, if you want, with a common get resource method. E.g.
```java
private <T> T getResource(Class<T> type) {
    return strimziResources.stream()
        .filter(type::isInstance)
        .map(type::cast)
        .findFirst()
        .orElseThrow(() -> new NoSuchElementException("No resource of type " + type));
}
```
Then all the other methods become simply the builder:
```java
private ServiceAccount getBundleServiceAccount() {
    return new ServiceAccountBuilder(getResource(ServiceAccount.class))
        .editMetadata()
            .withNamespace(this.deploymentNamespace)
        .endMetadata()
        .build();
}
```

MikeEdgar (approved):

systemtests/src/main/java/com/github/streamshub/systemtests/setup/StrimziOperatorSetup.java
jankalinic: Sounds great. Thanks

systemtests/src/main/java/com/github/streamshub/systemtests/setup/StrimziOperatorSetup.java
jankalinic: Thank you, don't know how i missed that.

im-konge (commented):
LGTM, just few suggestions or recommendations.

systemtests/src/main/java/com/github/streamshub/systemtests/setup/ConsoleOperatorSetup.java
```diff
@@ -0,0 +1,261 @@
...
+
+    public void setup() {
+        LOGGER.info("----------- Setup Console Operator -----------");
+
+        if (!ResourceUtils.listKubeResourcesByPrefix(Deployment.class, deploymentNamespace, operatorDeploymentName).isEmpty()) {
+            LOGGER.warn("Console Operator is already deployed. Skipping deployment");
+            return;
+        }
+
+        if (SetupUtils.isOlmInstall()) {
```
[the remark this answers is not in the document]
jankalinic: I moved isOlmInstall to Environment and use InstallConfig

systemtests/src/main/java/com/github/streamshub/systemtests/setup/ConsoleOperatorSetup.java
```diff
@@ -0,0 +1,261 @@
...
+            LOGGER.info("Setup Console Operator using OLM");
+            // Check for non OpenShift clusters
+            if (!ClusterUtils.isOcp() &&
+                ResourceUtils.getKubeResource(CustomResourceDefinition.class, "subscriptions.operators.coreos.com") == null) {
+                throw new OperatorSdkNotInstalledException("Operator SDK is not installed on the current cluster. Cannot install Console Operator using subscriptions");
+            }
+            KubeResourceManager.get().createResourceWithWait(getOlmOperatorGroup());
+            KubeResourceManager.get().createResourceWithWait(getOlmSubscription());
+            WaitUtils.waitForDeploymentWithPrefixIsReady(deploymentNamespace, operatorDeploymentName);
```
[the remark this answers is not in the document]
jankalinic: I'll leave it to next PRs. This PR is big enough for reviews. Good idea for testing though!

systemtests/src/main/java/com/github/streamshub/systemtests/setup/StrimziOperatorSetup.java
```diff
@@ -139,27 +120,37 @@ private ClusterRoleBinding[] getStrimziWatchAllCrbs() {
...
 
     private ConfigMap getBundleConfigMap() {
-        return new ConfigMapBuilder(TestFrameUtils.configFromYaml(strimziResources.stream()
-            .filter(file -> file.getName().startsWith("050-ConfigMap")).toList().get(0), ConfigMap.class))
-            .editMetadata()
-                .withNamespace(this.deploymentNamespace)
-                .withName(this.deploymentName)
-            .endMetadata()
-            .build();
+        return strimziResources.stream()
```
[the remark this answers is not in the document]
jankalinic: Using Strimzi with HELM, it made the setup cleaner. Thanks for the tip!

jankalinic:
Made big improvements in terms of installation of Strimzi -> now installed with HELM. Also Console now uses config to install OLM and Yaml separately, making it more flexible and possible to add new install type in the future.

im-konge (approved):
LGTM, thanks :)

systemtests/src/main/java/com/github/streamshub/systemtests/Environment.java
```diff
@@ -10,38 +12,55 @@
 public class Environment {
     private static final TestEnvironmentVariables ENVS = new TestEnvironmentVariables();
     // ------------------------------------------------------------------------------------------------------------------------------------------------------------
-    public static final String BUILD_ID = ENVS.getOrDefault("BUILD_ID", "0");
+
     public static final String CLIENT_TYPE = ENVS.getOrDefault("CLIENT_TYPE", "kubectl");
-    public static final String BROWSER_TYPE = ENVS.getOrDefault("BROWSER_TYPE", BrowserTypes.CHROMIUM.toString());
-    public static final String CONSOLE_DEPLOYMENT_NAME = ENVS.getOrDefault("CONSOLE_DEPLOYMENT_NAME", "streamshub-console");
+
+
```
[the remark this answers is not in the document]
jankalinic: Right, i need to implement some checkstyle to avoid those. @MikeEdgar Shall we do that for the main pom?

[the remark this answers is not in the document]
MikeEdgar: The main pom already has checkstyle, so maybe it's just something to add to the `.checkstyle/checkstyle.xml` configuration.

MikeEdgar (approved):

PR #1601, opened 2025-04-03
Include node pool size (count) in summary meta object

MikeEdgar:


ui/api/nodes/schema.ts
```diff
@@ -19,7 +19,10 @@ const ControllerStatusSchema = z.union([
   z.literal("Unknown"),
 ]);
 
-const NodePoolsSchema = z.record(z.string(), z.array(z.string()));
+const NodePoolsSchema = z.record(z.string(), z.object({
+  roles: z.array(z.string()),
+  count: z.number(),
+}));
```
MikeEdgar, line 25: @hemahg - as discussed. Let me know if this isn't enough/correct for the node pool filter.

hemahg (approved):
yes this data is enough. Thank you @MikeEdgar

PR #1603, opened 2025-04-04
Add OIDC/OAuth2 support for Prometheus and Apicurio clients

### LINKED ISSUE #1535: [Improvement] Add OAuth support for schema registry
Hi,

I am using the console with our AMQ Streams Kafka in OpenShift. I am using Apicurio for the schema registry. For the moment, as the message are serialized, it's not possible de read the topic by the console.

Is this planned to add support for a schema registry for deconding message?

Image

Thanks

MikeEdgar:
Closes #1535 

With this change, users may use Basic, Bearer token, or OIDC/OAuth authentication to communicate with Prometheus and Apicurio Registry.

For example, the `oidc` authentication for a schema registry may be configured like below. Note, that this allows the console to obtain tokens from Keycloak for use with Apicurio. Both the OIDC authentication server and the schema registry may require a trust store (shown below) when either or both use self-signed certificates.

Note, this configuration is the console's YAML file, not the CR. The CR looks almost the same, but the references to secrets and truststores would instead reference kube `ConfigMap`s or `Secret`s.

```yaml
schemaRegistries:
  - name: test-registry
    url: http://apicurio.example.com/api/v2
    authentication:
      oidc:
        authServerUrl: https://keycloak.example.com/realms/my-realm
        clientId: my-client
        clientSecret:
          value: my-client-secret
        method: POST
        grantType: CLIENT
        trustStore: # trust store for keycloak communication
          type: JKS
          content:
            valueFrom: /path/to/my/truststore.jks
    trustStore: # trust store for Apicurio communication
      type: PEM
      content:
        valueFrom: /path/to/my/cert-chain.pem
```


jankalinic (approved):
Thank you! To me looks good.

PR #1616, opened 2025-04-11
Add Console instance and Kafka deployment  + test case configuration

jankalinic:
This PR enables Console instance deployment and Kafka setup required for tests.
* Added `TestCaseConfig` which helps store test related data in an object instead of many variables for each test. It also handles playwright context, so calling playwright should be easier with playwright context being part of this class.
* Added env for minikube setup
* Refactored ExampleFIlePaths, made all strings actual Paths as they are anyway used in this way to get YAML files.
* Added Kafka types, they help check the actual ready state of the kafka components that is not provided by the testframe.
* Added Console instance setup
* Added PlaywrightUtils for browser creation.
* Added Utils class for utilities which i feel like will not belong to any of the existing utils classes. In this iteration it is hashStub function, which helps identify components belonging to specific testcase / namespace / kafka cluster
* Added readiness check for the Secrets, as i could not find anything in Readiness class or any related ones.
* Modified AbstractST to delete resources with ResourceManager from test frame by the sequential method. This proved to be a better way than async as some components are trying to delete a resource that is already deleted (HELM to be specific).
* And finally added FakeST class to show how the test case setup could look like while there are no regular tests yet (can be removed, but it helps to see how the setup is working).

Tested on openshift [routes] and minikube[ingress](linux only).


systemtests/pom.xml
```diff
@@ -176,7 +176,7 @@
                     <execution>
                         <id>system-tests</id>
                         <goals>
-                            <goal>verify</goal>
+                            <goal>integration-test</goal>
```
MikeEdgar, line 179: We need both right?

```suggestion
                            <goal>integration-test</goal>
                            <goal>verify</goal>
```

systemtests/src/main/java/com/github/streamshub/systemtests/constants/Constants.java
```diff
@@ -1,14 +1,18 @@
...
 public class Constants {
     private Constants() {}
 
     // --------------------------------
     // ------------ General -----------
     // --------------------------------
+    /**
+     * Cluster
+     */
+    public static final String KUBE_DNS = "192.168.49.2.nip.io";
```
MikeEdgar, line 12: Is this IP fixed?

systemtests/src/main/java/com/github/streamshub/systemtests/resourcetypes/ConsoleType.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public NonNamespaceOperation<?, ?, ?> getClient() {
+        return consoleClient();
+    }
+
+    @Override
+    public String getKind() {
+        return Console.class.getSimpleName();
```
MikeEdgar, line 28: Might use `HasMetadata.getKind` here.

systemtests/src/main/java/com/github/streamshub/systemtests/resourcetypes/ConsoleType.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public void create(Console console) {
+        consoleClient().inNamespace(console.getMetadata().getNamespace()).resource(console).create();
```
MikeEdgar, line 33: I believe if you use `.resource` it can find the namespace, which makes `.inNamespace` redundant since it's just taking the NS from the same object.

systemtests/src/main/java/com/github/streamshub/systemtests/utils/ConsoleUtils.java
```diff
@@ -0,0 +1,15 @@
+package com.github.streamshub.systemtests.utils;
+
+import io.fabric8.openshift.api.model.Route;
+
+public class ConsoleUtils {
+    private ConsoleUtils() {}
+
+    public static String getConsoleUiUrl(String namespace, String consoleInstanceName, boolean httpSecure) {
```
MikeEdgar, line 8: We should put this in the Console CR's `.status` in the future.

systemtests/src/main/java/com/github/streamshub/systemtests/constants/Constants.java
jankalinic: I believe for minikube it is, shall i add an ENV for it to make it more flexible?

systemtests/src/main/java/com/github/streamshub/systemtests/resourcetypes/ConsoleType.java
jankalinic: Right, i think there were some cases in which it did not work, but i tried it for our resources and it seems fine without it. Thanks for noticing it.

systemtests/src/main/java/com/github/streamshub/systemtests/TestCaseConfig.java
```diff
@@ -0,0 +1,108 @@
...
+    public void defaultTestCaseSetup() {
+        // Namespace
+        createNamespaceIfNeeded();
+        // Kafka
+        addKafka(new KafkaSetupConfig(namespaceName));
+        setupKafkaClustersIfNeeded();
+        // Console
+        setConsoleInstance(new ConsoleInstanceSetup(namespaceName, getDefaultKafkaCluster().getClusterName()));
+        consoleInstanceSetup.deploy();
+    }
```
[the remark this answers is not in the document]
jankalinic: I get what you're saying, my idea behind this was to config a kafkaSetup, then pass it to the testcase config. This default method is purely for default purposes which i believe will be anyway the main target / most used config. I could remove the default setup here to make it more verbose while setting up a namespaces then kafka then console instance.

systemtests/src/main/java/com/github/streamshub/systemtests/setup/console/ConsoleInstanceSetup.java
```diff
@@ -0,0 +1,86 @@
...
+        if (ResourceUtils.getKubeResource(Deployment.class, deploymentNamespace, instanceName) != null) {
+            LOGGER.warn("Skipping Console Instance deployment. It is already deployed");
+            return;
+        }
+
+        KubeResourceManager.get().createResourceWithWait(getExampleConsoleInstance());
+        LOGGER.info("Console deployed and available at {}", ConsoleUtils.getConsoleUiUrl(deploymentNamespace, instanceName, true));
+    }
+
+    private Console getExampleConsoleInstance() {
```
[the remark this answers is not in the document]
jankalinic: So do you suggest leaving out the ConsoleInstanceSetup class completely and setup console instance in each test after kafka is deployed?

systemtests/src/test/java/com/github/streamshub/systemtests/system/FakeST.java
```diff
@@ -0,0 +1,28 @@
...
+        // Instantiate test case config for current context
+        TestCaseConfig tcc = new TestCaseConfig(KubeResourceManager.get().getTestContext(), NAMESPACE_PREFIX);
+        // Create namespace, deploy kafka, deploy console instance
+        tcc.defaultTestCaseSetup();
+
+        LOGGER.info("Test starts now");
+        assertTrue(true);
+
+        // Close current testcase playwright context
+        tcc.close();
```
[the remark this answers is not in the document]
jankalinic: I wanted to use Playwright fixtures and not care about it at all, but it's in experimental state and it does not allow selecting different browsers for the test. So in case you run tests in headless mode turned off, the browser window would hang open there. I guess i could store it in context, would you suggest using the same system strimzi had - storagemap protected in the AbstractST class that got those things initialized in before each and cleaned in after each?

systemtests/src/main/java/com/github/streamshub/systemtests/setup/console/ConsoleInstanceSetup.java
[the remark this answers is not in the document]
im-konge: Yeah that's my suggestion. You can create some methods that will return something default (applied to every test case or every deployment of Console). But it's better to have these things visible in case that someone else (than you) will want to write the tests. Especially for newcomers it's really confusing to see what is happening where. That's why I'm suggesting to have everything visible in the test case rather than having it hidden somewhere else. Because if I'm someone new and I will look at the test case, I'm wondering what is the magic that Kafka, Console and everything else is deployed in this particular test case.

Just speaking of my own experience.

systemtests/src/main/java/com/github/streamshub/systemtests/TestCaseConfig.java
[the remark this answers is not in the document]
im-konge: For the Kafka part, I think that it's fine that you have all things "generated" in terms of the fields, names etc. and the `TestCaseConfig` holds everything related. I would just move this part out of the config and keep all the `Config` classes as `records` for storing the stuff. (obviously you don't have to implement `record` but I wouldn't do anything else here).

.github/workflows/integration.yml
```diff
@@ -187,7 +187,7 @@ jobs:
           #
           # See: https://quarkus.io/guides/tests-with-coverage#coverage-for-integration-tests
           #
-          mvn verify -B --no-transfer-progress \
+          mvn verify -B --no-transfer-progress -pl common,api,operator\
```
im-konge, line 190: Why is this needed?

jankalinic: The integration runs API and OPERATOR modules, but with the verify and integration in the pom.xml it runs systemtests as well. I thought i should not disable STs by adding a default profile that runs none tests. Could maybe do that using a profile. WDYT? @MikeEdgar

im-konge: You can do it differently. You can disable running the STs for specific goal and then use something else (like mvn clean install vs mvn verify). Or you can create property `skipSTs` and its value can be set for whatever triggers the tests. For example we have `<it.skip>true</it.skip>` and that is the property for Failsafe (or is it Surefire 🤔 ) for skipping the tests. You can play with it a bit. I wouldn't do this, because every time when there will be new module, you will need to remember that you have to add it to this list of modules where you want to run UTs and ITs (and it's easily forgettable).

jankalinic: I found it more fitting in this case, but true could skip it here with skip variable i anyway use.

systemtests/src/main/java/com/github/streamshub/systemtests/TestCaseConfig.java
```diff
@@ -0,0 +1,49 @@
...
+    private final String testName;
+    private final String namespace;
+    private final Playwright playwright;
+    private final Browser browser;
+    private final BrowserContext context;
+    private final Page page;
+
+    public TestCaseConfig(ExtensionContext extensionContext) {
+        this.testName = extensionContext.getTestMethod().isPresent() ? extensionContext.getTestMethod().get().getName() : extensionContext.getTestClass().get().getSimpleName();
+        this.namespace = extensionContext.getTestClass().isPresent() ?  extensionContext.getTestClass().get().getSimpleName().toLowerCase(Locale.ENGLISH) + "-" + Utils.hashStub(testName) : "nullClass";
```
MikeEdgar, line 24: A few adjustments to the `Optional` handling to clean this up and make Sonar happy:

```suggestion
        this.testName = extensionContext.getTestMethod().map(Method::getName)
                .or(() -> extensionContext.getTestClass().map(Class::getSimpleName))
                .orElseThrow();
        this.namespace = extensionContext.getTestClass()
                .map(Class::getSimpleName)
                .map(name -> name.toLowerCase(Locale.ENGLISH) + "-" + Utils.hashStub(testName))
                .orElse("nullClass");
```

jankalinic:
@im-konge @MikeEdgar  can we merge this PR? or are there any other things you would like to discuss / change

MikeEdgar (approved):
Nothing more from me on this. Thanks @jankalinic

im-konge (approved):

Issue #1526, opened 2025-03-12
Submit bundles to OperatorHub community operator catalogs


MikeEdgar:
Create a script to generate the catalog layout for the community operators repositories using the OLM metadata from operator/src/main/olm. Since community operators do not use file-based catalogs (FBC), we will need to extract the bundle metadata from the released bundle images and patch it accordingly. This gives us a more maintainable process than manually editing the manifests for new submissions to the catalog repositories.

- https://github.com/k8s-operatorhub/community-operators
- https://github.com/redhat-openshift-ecosystem/community-operators-prod

tomncooper:
When this is complete we should enable the instructions in the docs deploytment assembly: `docs/sources/assemblies/assembly-deploying.adoc`

tomncooper:
We should aim to do this ASAP but we should check to see if the RH and the Community have aligned on their format.

PR #1647, opened 2025-05-08
Inital documentation commit

tomncooper:
This PR contains documentation originally written in a separate repository by @PaulRMellor and @MikeEdgar.

We are moving it here so it can be maintained in concert with the codebase and this will be pulled into the main [StreamsHub Site](https://github.com/streamshub/streamshub-site). The current main branch docs will always be pulled in and each tagged version from now on will have an independent space on the main StreamsHub site.


PaulRMellor (approved):
Look good.
Thanks, Tom

docs/sources/attributes.adoc
```diff
@@ -0,0 +1,21 @@
...
+:numbered:
+:sectanchors!:
+:sectnums:
+:source-highlighter: highlightjs
+:toc: left
+:linkattrs:
+:toclevels: 4
+
+// Variables
+:minKubernetesVersion: 1.19
```
tomncooper, line 18: @MikeEdgar What should this value be?

tomncooper: Is that the minimum, the console won't work below that version?

tomncooper:
Thanks both, one thing I forgot to mention (which in hindsight is pretty important). We mention the StreamsHub Console Operator a lot here and AFAIK that is not actually released on the OperatorHub yet? 

I can leave it in for now or comment those deployment pages out until we release it on the OH?

MikeEdgar:
It would probably be OK to comment them out and add a note to #1526 to re-enable them when completing that task.

tomncooper:
Ok I disabled the OLM instructions in the deployment assembly (the source for them is still there it just won't render by default). The deployment intro now reads:
```asciidoc
* By applying a `Console` Custom Resource Definition (CRD)
* Once the StreamsHub operator is submitted to the OperatorHub (see https://github.com/streamshub/console/issues/1526[Issue 1526] for details), you can install the operator:
** By using the Operator Lifecycle Manager (OLM) command line interface (CLI)
** From the OperatorHub in the OpenShift web console (Openshift Clusters only)
```

PaulRMellor:
I was going to suggest a note in the readme as an alternative to commenting out

PR #1657, opened 2025-05-13
feat: Add pagination to list of kafka cluster

hemahg:
feat: Add pagination to list of kafka cluster


MikeEdgar (approved):
LGTM. Unrelated to this change, I noticed that the sort logic is incorrect in `api/api.ts` in function `sortParam`.

Instead of
```ts
return (order === "asc" ? "-" : "") + sortField;
```

It should be 
```ts
return (order === "desc" ? "-" : "") + sortField;
```

PR #1664, opened 2025-05-16
 Add support for filtering Kafka clusters by name

hemahg:
Add support for filtering Kafka clusters by name

### DIFF
--- a/api/src/main/java/com/github/streamshub/console/api/model/KafkaClusterFilterParams.java
+++ b/api/src/main/java/com/github/streamshub/console/api/model/KafkaClusterFilterParams.java
@@ -17,24 +17,24 @@ import io.xlate.validation.constraints.Expression;
 
 public class KafkaClusterFilterParams {
 
-  @QueryParam("filter[name]")
-  @Parameter(
-      description = "Retrieve only kafka clusters with a name matching this parameter",
-      schema = @Schema(implementation = String[].class, minItems = 2),
-      explode = Explode.FALSE)
-  @Expression(
-      when = "self != null",
-      value = "self.operator == 'eq' || self.operator == 'in' || self.operator == 'like'",
-      message = "unsupported filter operator, supported values: [ 'eq', 'in', 'like' ]",
-      payload = ErrorCategory.InvalidQueryParameter.class,
-      node = "filter[name]")
-  @Expression(
-      when = "self != null",
-      value = "self.operands.size() >= 1",
-      message = "at least 1 operand is required",
-      payload = ErrorCategory.InvalidQueryParameter.class,
-      node = "filter[name]")
-  FetchFilter nameFilter;
+    @QueryParam("filter[name]")
+    @Parameter(
+        description = "Retrieve only kafka clusters with a name matching this parameter",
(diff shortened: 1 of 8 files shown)


MikeEdgar:
@hemahg I forgot to mention you will need to add a `.filter(listSupport)` before the security check [1] to make use of the filter parameters.

[1] https://github.com/streamshub/console/blob/8ff75f7755953bc88cac7caafd2dfcfd5a6fc14e/api/src/main/java/com/github/streamshub/console/api/service/KafkaClusterService.java#L139

api/src/main/java/com/github/streamshub/console/api/model/KafkaClusterFilterParams.java
```diff
@@ -0,0 +1,48 @@
...
+      message = "unsupported filter operator, supported values: [ 'eq', 'in', 'like' ]",
+      payload = ErrorCategory.InvalidQueryParameter.class,
+      node = "filter[name]")
+  @Expression(
+      when = "self != null",
+      value = "self.operands.size() >= 1",
+      message = "at least 1 operand is required",
+      payload = ErrorCategory.InvalidQueryParameter.class,
+      node = "filter[name]")
+  FetchFilter nameFilter;
```
MikeEdgar, line 37: This should resolve the checkstyle violation:

```suggestion
    @QueryParam("filter[name]")
    @Parameter(
        description = "Retrieve only kafka clusters with a name matching this parameter",
        schema = @Schema(implementation = String[].class, minItems = 2),
        explode = Explode.FALSE)
    @Expression(
        when = "self != null",
        value = "self.operator == 'eq' || self.operator == 'in' || self.operator == 'like'",
        message = "unsupported filter operator, supported values: [ 'eq', 'in', 'like' ]",
        payload = ErrorCategory.InvalidQueryParameter.class,
        node = "filter[name]")
    @Expression(
        when = "self != null",
        value = "self.operands.size() >= 1",
        message = "at least 1 operand is required",
        payload = ErrorCategory.InvalidQueryParameter.class,
        node = "filter[name]")
    FetchFilter nameFilter;
```

MikeEdgar:
Take a look at `KafkaClustersResourceIT`. There is an existing test `testListClusters` that can be modified to pass the new `filter[name]` query parameter. There should be 3 Kafka's by default, `test-kafka1`, `test-kafka2` and `test-kafkaY`. The test only needs to include the first assertion about `data.attributes.name` checking the names of the ones matching the filter you provide.

MikeEdgar (approved):
Looks good. I think we should add something to prevent redirecting to the login page when the search results in finding only one Kafka. It makes sense if only one cluster is configured in the console, but feels unusual if the search finds one result. Maybe something like this?

```
  if (allClusters?.data.length === 1 && !oidcEnabled) {
    let hasSearch: boolean = Object.values(searchParams).some(p => p !== undefined);
    if (!hasSearch) {
      return <RedirectOnLoad url={`/kafka/${allClusters.data[0].id}/login`} />;
    }
  }
```

hemahg:
Good catch,I didn't noticed it. fixed it

MikeEdgar (approved):
:+1:

PR #1673, opened 2025-05-20
Fix the gap issue in the create topic

hemahg:
Fix the gap issue in the create topic

Screenshot 2025-05-20 at 6 41 25 PM


MikeEdgar (approved):
Thanks @hemahg !

PR #1679, opened 2025-05-21
fix: properly implement topic mutate operations for KafkaTopic CRs

MikeEdgar:
The operations to create, update, and delete topics did not consider whether a Kafka cluster is using the Strimzi Topic Operator (for creation) or whether an existing topic is backed by a Strimzi `KafkaTopic` CR. This change addresses the gap by:

- create a `KafkaTopic` CR during topic creation if the topic entity operator is enabled for a Kafka CR. Otherwise, create directly in Kafka
- update the `KafkaTopic` CR for managed topics, else directly in Kafka
- delete the `KafkaTopic` CR for managed topics, else directly in Kafka

This functionality is not accessible in the UI when read-only mode is enabled.


MikeEdgar:
Additional integration tests still needed.

PR #1696, opened 2025-05-29
fix: choose node icon based on aggregate status

MikeEdgar:
Also suppress non-PEM truststore warning when no truststore is configured at all.

With OK statuses:
image

With warning statuses:
image


jankalinic (approved):
Thanks

PR #1720, opened 2025-06-04
[ST] Add TopicST tests

jankalinic:


systemtests/pom.xml
```diff
@@ -27,7 +41,7 @@
         <mockito.version>5.15.2</mockito.version>
         <powermock.version>2.0.9</powermock.version>
         <!-- Used for test-frame -->
-        <test-frame.version>0.13.0</test-frame.version>
+        <test-frame.version>0.14.0-SNAPSHOT</test-frame.version>
```
im-konge, line 44: You can use `0.14.1` now and remove the snapshot repository

systemtests/src/main/java/com/github/streamshub/systemtests/utils/resourceutils/KafkaClientsUtils.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    private static String getScramShaConfig(String namespace, String userName, SecurityProtocol securityProtocol) {
+        final String saslJaasConfigDecrypted = Utils.decodeFromBase64(ResourceUtils.getKubeResource(Secret.class, namespace, userName).getData().get("sasl.jaas.config"));
+        return  "sasl.mechanism=SCRAM-SHA-512\n" +
+            "security.protocol=" + securityProtocol + "\n" +
+            "sasl.jaas.config=" + saslJaasConfigDecrypted + "\n";
+    }
+
+    public static KafkaClientsBuilder plainClientBuilder(String namespace, String topicName, String producer, String consumer, String bootstrap, int messageCount, int delayMs) {
```
im-konge, line 31: Why is it better to have a method containing all these things instead of using the builder directly?
I mean, you can generate the consumer group in some method which you can use the builders.

systemtests/src/main/java/com/github/streamshub/systemtests/utils/resourceutils/KafkaClientsUtils.java
```diff
@@ -0,0 +1,44 @@
...
+        String uid = UUID.randomUUID().toString().substring(0, 4);
+        return new KafkaClientsBuilder()
+            .withNamespaceName(namespace)
+            .withTopicName(topicName)
+            .withMessageCount(messageCount)
+            .withDelayMs(delayMs)
+            .withProducerName(producer)
+            .withConsumerName(consumer)
+            .withConsumerGroup(consumer + "-group-" + uid)
+            .withAdditionalConfig("group.instance.id=group-" + uid + "\n")
```
im-konge, line 41: Just out of curiosity, you need the unique id for the particular consumer for some tests?

systemtests/src/main/java/com/github/streamshub/systemtests/utils/resourceutils/KafkaTopicUtils.java
```diff
@@ -0,0 +1,201 @@
...
+    
+    public static List<KafkaTopic> createTopicsAndReturn(String namespace, String kafkaName, String topicNamePrefix, int numberToCreate, boolean waitForTopics, int partitions, int replicas, int minIsr) {
+        LOGGER.info("Create {} topics for cluster {} with topic name prefix {}", numberToCreate, kafkaName, topicNamePrefix);
+
+        List<KafkaTopic> topics = IntStream.range(0, numberToCreate)
+            .mapToObj(i -> defaultTopic(namespace, kafkaName, topicNamePrefix + "-" + i, partitions, replicas, minIsr).build())
+            .toList();
+
+        if (waitForTopics) {
+            topics.forEach(KubeResourceManager.get()::createResourceWithWait);
```
im-konge, line 34: instead of for each, which will do this sequently, you can just pass the list to array and use the async method probably?

systemtests/src/main/java/com/github/streamshub/systemtests/utils/resourceutils/KafkaTopicUtils.java
```diff
@@ -0,0 +1,201 @@
...
+        } else {
+            topics.forEach(KubeResourceManager.get()::createResourceWithoutWait);
+        }
+        return topics;
+    }
+
+
+    public static List<KafkaTopic> createUnderReplicatedTopicsAndReturn(String namespace, String kafkaName, String kafkaUser, String topicNamePrefix, int numberToCreate, int messageCount, int partitions, int replicas, int minIsr) {
```
im-konge, line 42: That's a bit strange name for the actions it really does.
With `createUnderReplicatedTopicsAndReturn` I would imagine that it will create just the KafkaTopic CRs with some strange configuration. But you are scaling the nodePools up, waiting for scale up, then you are creating topics, sending messages  and then scaling down.

How many times you will use this? Shouldn't it be split into multiple methods?

systemtests/src/main/java/com/github/streamshub/systemtests/utils/resourceutils/KafkaTopicUtils.java
```diff
@@ -0,0 +1,201 @@
...
+                .endSpec()
+                .build());
+
+        WaitUtils.waitForPodsReady(namespace, Labels.getKnpBrokerLabelSelector(kafkaName), knp.getSpec().getReplicas() - 1, true);
+
+        KafkaUtils.removeAnnotation(namespace, kafkaName, ResourceAnnotations.ANNO_STRIMZI_IO_SKIP_BROKER_SCALEDOWN_CHECK, true);
+        return kafkaTopics;
+    }
+
+    public static List<KafkaTopic> createUnavailableTopicsAndReturn(String namespace, String kafkaName, String kafkaUser, String topicNamePrefix, int numberToCreate, int messageCount, int partitions, int replicas, int minIsr) {
```
im-konge, line 88: Same as above

systemtests/src/main/java/com/github/streamshub/systemtests/utils/resourceutils/KafkaTopicUtils.java
jankalinic: yea, sure, i do it on other places so i can unify this

systemtests/src/main/java/com/github/streamshub/systemtests/utils/resourceutils/KafkaClientsUtils.java
jankalinic: Not in particular, i just found it useful for debuging on existing kafka

systemtests/src/main/java/com/github/streamshub/systemtests/utils/resourceutils/KafkaTopicUtils.java
jankalinic: It's used multiple times for a simple creation of udnerreplicated topics, i think you cannot create underreplicated topic with plain CR so the name made perfect sense for me it does create those topics and returns them for later usage.... maybe a more fiting name starting with `prepare` or `provision` maybe even `setup` ? i mean i do not want to repeat the whole scaling and moving partitions each test where it's used it would inflate all the tests method block

systemtests/src/main/java/com/github/streamshub/systemtests/utils/resourceutils/KafkaClientsUtils.java
jankalinic: I think i can move it up one nest to get rid of the plain builder, but i'd still rather keep it short like this

```
KafkaClientsUtils.scramShaPlainTextClientBuilder(namespace, kafkaName, kafkaUser, kt.getMetadata().getName(), messageCount).build();
```
instead of using the whole builder
```
KafkaClientsBuilder()
            .withNamespaceName(namespace)
            .withTopicName(topicName)
            .withMessageCount(messageCount)
            .withDelayMs(0)
            .withProducerName(topicName + "-producer")
            .withConsumerName(topicName + "-consumer")
            .withConsumerGroup(topicName + "-consumer" + "-group-" + uid)
            .withBootstrapAddress(KafkaUtils.getPlainScramShaBootstrapAddress(kafkaName))
            .withUsername(kafkaUser)
            .withAdditionalConfig("group.instance.id=group-" + uid + "\n" + getScramShaConfig(namespace, kafkaUser, SecurityProtocol.SASL_PLAINTEXT));
            ```

im-konge: In that case you don't need to create builders. The main benefit of having builders is that you can layer the things that you want to set. If you hide it into some method, you are not using the main thing.

systemtests/src/main/java/com/github/streamshub/systemtests/utils/resourceutils/KafkaTopicUtils.java
im-konge: That's what I mean, it does a lot of other things that the name actually suggests. Plus, you are missing the Javadoc there, which would help also others to understand what is the method doing.
And if you have multiple tests checking the things with same setup of topics - in this case under replicated - I think you can have it in some before all and then run against it few tests? Because this will take a lot of time if you will have it in every test.

systemtests/src/main/java/com/github/streamshub/systemtests/utils/resourceutils/KafkaClientsUtils.java
jankalinic: I did not create the classes of builders, those are generated from KafkaClients. I think using default builder values could be useful to not repeat yourself if you know you need specific listener for multiple instances... You may argue then maybe not use builder at all and use KafkaClient constructor with all the parameters? I mean i wouldbe okay with that and remove it from Utils. And if there is a test where scramShaConfig isn't used i could layer clients like here with builders, that's why i layered it before with plainBuilder->plainScramShaBuilder->built clients, to utilize the reason why builders are used.

MikeEdgar (commented):
A few suggestions for some of the code warnings.

PR #1722, opened 2025-06-13
Refactor JSON:API model classes

MikeEdgar:
Refactoring of JSON:API [1] models to separate package to ease upcoming development for Kafka Connect endpoints. There are no functional changes with this change.

[1] https://jsonapi.org/


MikeEdgar:
> At first I was thinking if the `record` can be used instead of classes for the `model/jsonapi` classes, but I guess you need the API validation, so it wouldn't be much useful in this case.

Yep, that as well as inheritance, which unfortunately isn't supported for records.

PR #1731, opened 2025-06-23
fix: correct list structures in security rule examples

MikeEdgar:
Remove incorrect `-` from security rules. Their presence results in additional list entries where each of the entries are missing one or the other require key, `privileges` or `resources`.


PaulRMellor (approved):
Thank, Mike. Looks good

PR #1753, opened 2025-07-09
Upgrade to latest version of Storybook

MikeEdgar:
Used provided tool to upgrade and run automatic migrations

All changes in this PR were made with the upgrade tool.
```
npx storybook@latest upgrade
```

### DIFF
--- a/ui/components/ClusterOverview/PageLayout.stories.tsx
+++ b/ui/components/ClusterOverview/PageLayout.stories.tsx
@@ -1,5 +1,5 @@
-import type { Meta, StoryObj } from "@storybook/react";
-import { composeStories } from "@storybook/react";
+import type { Meta, StoryObj } from "@storybook/nextjs";
+import { composeStories } from "@storybook/nextjs";
 import * as ClusterCardStories from "./ClusterCard.stories";
 import { PageLayout } from "./PageLayout";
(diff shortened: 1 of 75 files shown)


ui/components/ClusterOverview/PageLayout.stories.tsx
```diff
@@ -1,5 +1,5 @@
-import type { Meta, StoryObj } from "@storybook/nextjs";
-import { composeStories } from "@storybook/nextjs";
+import type { Meta, StoryObj } from "@storybook/react";
+import { composeStories } from "@storybook/react";
```
hemahg, line 2: The issue is that the story uses a context provider, but Next.js doesn’t provide native support for it. I tried using the Next.js Router addon, but it's not compatible with Next.js 14. The best solution I found was to replace the Next.js wrapper with a plain React setup. Let me know what you think about this approach.

hemahg: Storybook is a Dev dependence,it won’t affect production

MikeEdgar: Ah, I just realized the build passed with just this change. Do you think we can leave it as-is or is it better to use storybook/react everywhere instead of storybook/nextjs? I'm fine to just have a different approach for this one test if that makes sense.

PR #1737, opened 2025-06-26
add Kafka Connect REST interface and service

MikeEdgar:

PR #1754, opened 2025-07-09
Add support for multi-typed filters

MikeEdgar:
Currently, when the API accepts query parameters for filtering (e.g. `filter[name]`), the server will convert the filter into a Java `Predicate` with a generic type argument of the collection being filtered. For example, filters for the `/api/kafkas` collection will result in `Predicate<KafkaCluster>`.

This PR adds support for predicates against multiple types. This will be useful in PR #1737 where there is a need to filter not just by the primary collection type, but also by alternate types to preemptively limit the number of REST calls made to Kafka Connect endpoints. The changes here merely lay the groundwork to be used by that other PR and there are no functional changes to the API yet.


jankalinic (approved):
LGTM

PR #1777, opened 2025-07-17
fix: Remove wizard title as it was applied to all wizard steps

hemahg:
fix: Remove wizard title as it was applied to all wizard steps


hemahg:
I've removed the wizard steps from the tooltip content since they were being displayed as the tooltip text for the entire wizard. Also, we’re unable to add a tooltip directly to the "+" or "−" buttons because they are part of the NumberInput's internal implementation, and the component doesn’t expose direct control over those button elements.

Issue #1691, opened 2025-05-28
Make `groups` scope configurable


MikeEdgar:
Currently, the UI hard-codes a requirement that the OIDC provider supports a `groups` scope [1] and the API hard-codes the role claim path as `groups` [2]. Both of these should be configurable.

[1] https://github.com/streamshub/console/blob/403be5876df5db0e8e30f313333a4f2265a4dcf1/ui/app/api/auth/%5B...nextauth%5D/oidc.ts#L27
[2] https://github.com/streamshub/console/blob/403be5876df5db0e8e30f313333a4f2265a4dcf1/api/src/main/java/com/github/streamshub/console/api/security/OidcTenantConfigResolver.java#L47-L49

PR #1793, opened 2025-07-22
Make Scopes configurable

hemahg:
Make Scopes configurable 

fixes https://github.com/streamshub/console/issues/1691

### DIFF
--- a/common/src/main/java/com/github/streamshub/console/config/authentication/OIDC.java
+++ b/common/src/main/java/com/github/streamshub/console/config/authentication/OIDC.java
@@ -48,7 +48,6 @@ public class OIDC implements Trustable {
     private Boolean absoluteExpiresIn;
     private GrantType grantType = GrantType.CLIENT;
     private Map<String, String> grantOptions;
-    private List<String> roleClaimPath;
     private TrustStoreConfig trustStore;
 
     @Override
@@ -127,14 +126,6 @@ public class OIDC implements Trustable {
 
     public void setGrantOptions(Map<String, String> grantOptions) {
         this.grantOptions = grantOptions;
-    }
-
-    public List<String> getRoleClaimPath() {
-        return roleClaimPath;
-    }
-
-    public void setRoleClaimPath(List<String> roleClaimPath) {
-        this.roleClaimPath = roleClaimPath;
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


common/src/main/java/com/github/streamshub/console/config/security/OidcConfig.java
```diff
@@ -22,6 +24,8 @@ public class OidcConfig implements Trustable {
     private String clientId;
     @NotBlank
     private String clientSecret;
+    @NotBlank
+    private List<String> roleClaimPath = List.of("groups");
```
MikeEdgar, line 28: We can actually drop the default of `"groups"`, it's already assumed by the Quarkus extension as the claim path for the roles.

```suggestion
    private List<@NotBlank String> roleClaimPath; 
```

MikeEdgar (commented):
This looks good, thanks @hemahg . I missed something with the validation annotations in the suggestion about changing to a list.

I also forgot earlier that these two new configuration options are needed in the `Console` CR's class `com.github.streamshub.console.api.v1alpha1.spec.security.Oidc` in the `operator` module and copied from the CR to the config YAML model here: https://github.com/streamshub/console/blob/c1b8c771917f9961a53b339a46ad7f26025f78e4/operator/src/main/java/com/github/streamshub/console/dependents/ConfigurationProcessor.java#L430-L436

common/src/main/java/com/github/streamshub/console/config/security/OidcConfig.java
hemahg: Done

hemahg:
working on review comment https://github.com/streamshub/console/pull/1793#pullrequestreview-3048537493

operator/src/main/java/com/github/streamshub/console/api/v1alpha1/spec/authentication/OIDC.java
MikeEdgar, line 1: Same as above, no change needed here.

common/src/main/java/com/github/streamshub/console/config/authentication/OIDC.java
MikeEdgar, line 1: No change needed to this configuration object. This one is used for OIDC clients that are reaching out to connect to Apicurio, Prometheus, etc. `OidcConfig` is the only place where we need to know about the location of roles within the token. That's the configuration that console uses to secure itself.

MikeEdgar (changes requested):

MikeEdgar (approved):
LGTM, thanks @hemahg !

PR #1821, opened 2025-07-31
fix: use correct class in setup of connector name filter

MikeEdgar:


hemahg (approved):
Thanks for the quick fix

Issue #1838, opened 2025-08-07
Kafka nodes tab in console: not authorized exception when using Oidc authentication.


btison:
Issue when using Oauth authentication for the console: Not authorized exception when accessing the  `Kafka nodes` tab.

Security rules defined as follows:
```
    security:
      subjects: []
      roles:
      - name: "admins"
        rules:
        - resources:
          - "topics"
          - "topics/records"
          - "consumerGroups"
          - "rebalances"
          - "nodes/configs"
          resourceNames: []
          privileges:
          - "ALL"
      audit: []
```
The root cause seems to be in the way permissions are checked.
The permission that is checked when the nodes endpoint on the API is called is `console:kafkas/kafka-scram/nodes:[]:[LIST]` . Based on the configuration above, the assigned permissions are `[console:kafkas:[*]:[ALL], console:kafkas/kafka-scram/topics:[]:[ALL], console:kafkas/kafka-scram/topics/records:[]:[ALL], console:kafkas/kafka-scram/consumerGroups:[]:[ALL], console:kafkas/kafka-scram/rebalances:[]:[ALL], console:kafkas/kafka-scram/nodes/configs:[]:[ALL]]`.
In the `resourceDenied` method of the `ConsolePermission` class, the resource of the permission is checked using `equals()` against the resource of the assigned permissions, and so there is no match for `kafkas/kafka-scram/nodes` , and as a result access to the resource is denied.
Adding `nodes` to the resources in the rules fails on deployment as `nodes` is not a recognized resource type.

Potential fix:
Add `nodes` as recognized resource type in `com.github.streamshub.console.config.security.ResourceTypes`

**To Reproduce**
1. Install Kafka console (version 8.50 or higher) on OpenShift with OAuth authentication and security rules as defined above.
2. Open The console UI, go to the `Kafka nodes` tab
3. Expect to see the node information, instead a Not authorized exception is shown

PR #1839, opened 2025-08-07
fix: add 'nodes' to Resource types (#1838)

btison:


MikeEdgar (approved):
Thanks @btison , works as expected.

PR #1851, opened 2025-08-11
Include MirrorMaker type in connector type, add filter

MikeEdgar:
Add the MirrorMaker type to the connector and plugin types when the class of the connector or plugin indicates MirrorMaker (may be overridden via environment variable in cases where a customized class is used). The type may also be filtered so that only MM connectors or only non-MM connectors are returned.

Also validate Kafka Connect name/namespace/mirrorMaker flag combination is unique.


MikeEdgar:
@hemahg given the following possible `type`s:
- `sink`
- `source`
- `source:mm`
- `source:mm-checkpoint`
- `source:mm-heartbeat`

You can retrieve only mirror maker connectors with `filter[type]=in,source:mm,source:mm-checkpoint,source:mm-heartbeat` and only non-mirror maker connectors with `filter[type]=in,sink,source`.

PR #1856, opened 2025-08-12
Improve playwright tests, reduce flakes

MikeEdgar:
- Correct topic configuration `ariaLabel`
- Remove superfluous nav to overview page in fixture
- Attempt to improve use of locators and expectations


jankalinic (approved):
Seems like tests are passing fine, LGTM

PR #1857, opened 2025-08-13
Fix: Reduce extra space under Disk Space card in metrics charts

hemahg:
Reduce extra space under Disk Space card in metrics charts

### DIFF
--- a/ui/components/ClusterOverview/components/ChartDiskUsage.tsx
+++ b/ui/components/ClusterOverview/components/ChartDiskUsage.tsx
@@ -46,7 +46,11 @@ export function ChartDiskUsage({ usages, available }: ChartDiskUsageProps) {
     );
   }
   const CursorVoronoiContainer = createContainer("voronoi", "cursor");
-  const legendData: { name: string, childName: string, symbol?: { type: string } }[] = [];
+  const legendData: {
+    name: string;
+    childName: string;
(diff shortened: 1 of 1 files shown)


jankalinic:
@streamshub-bot help

jankalinic:
it now shows the same gap for each of them

image


but also i see the overflow is not hidden for the kafka card, so it became scrollable

image

hemahg:
Also, since the overview page is quite long, I don't see any visual issue caused by the scroll being enabled. So, having scroll enabled doesn't seem to negatively affect the layout in this case.

hemahg:
I removed the <div ref={containerRef} style={{ maxHeight: "280px" }}> wrapper. Adding maxHeight was unnecessarily restricting the height of the chart container, which caused overflow and an unwanted scrollbar to appear in the Kafka cluster details page.

jankalinic:
@streamshub-bot run tests --cluster-type=ocp --cluster-version=4.19 --profile=regression --test-case=ConsumerST --retry-count=1

jankalinic:
Thanks, that could help, i'll let the bot build the images and deploy latest changes to see it on cluster

jankalinic:
i believe the maxHeight isn't the issue, since it only worsened the memory usage card and the info card is still scrollable


image

Issue #633, opened 2024-04-02
UI: clicking on a consumer group with a dot in the name gives 404


btison:
**Describe the bug**
There is a consumer group with a dot in the name. When you click on the link to show the details of the consumer group, a 404 page is returned.

**To Reproduce**
1. Create a consumer group with a dot in the name (`globex.recommendation` in my example)
2. In the console, go to the consumer group overviewpage, and click on the link to the consumer group details
3. You get a 404 error page

**Expected behavior**
When clicking on the link of the consumer group, the details page should be shown.

Possible cause of the 404: in the url of the consumer group details page, the dot is not encoded. When encoding the dot in the browser address bar (`.` -> `%2e`), the link works correctly (see screenshots).

**Desktop (please complete the following information):**
 - OS: Fedora 39
 - Browser: Firefox
 - Version 122

![amq-console-4](https://github.com/eyefloaters/console/assets/568849/51daa2f2-6239-4588-a418-20baa14f5724)
![amq-console-5](https://github.com/eyefloaters/console/assets/568849/07b26e04-5bbf-4fe6-8fb7-7b88cbe1c44f)
![amq-console-6](https://github.com/eyefloaters/console/assets/568849/e9e4272a-ee2f-4bea-beaa-eedd752bcfee)

PR #1889, opened 2025-08-25
fix: always pass consumer group identifier through common encoding

MikeEdgar:
Relates to #633 and #634


hemahg:
I just saw the issue,the encodingURL will not works for dot as it treats dot as a valid character,I Came across a similar situation in Kafka connect where replace also didn’t work either

hemahg:
there are 2 investigation paths:
Check if there's a middleware for next.js router to handle .
Encode the ID in a different format (perhaps base64)

MikeEdgar:
> there are 2 investigation paths: Check if there's a middleware for next.js router to handle . Encode the ID in a different format (perhaps base64)

I'm debugging this also :laughing: It looks like it needs to be encoded/decoded in several places to properly display and also work with the backend API calls.

PR #1909, opened 2025-09-02
Add OIDC sign-out support using end_session_endpoint

hemahg:
Add OIDC sign-out support using end_session_endpoint


MikeEdgar (approved):
Looks good, thank you @hemahg. I'm no longer seeing any errors/warnings and it is working as expected for me.

PR #1963, opened 2025-09-15
Apply known user authorizations to UI (disable buttons, links, etc.)

MikeEdgar:


MikeEdgar:
Thanks @jankalinic , no additional changes for this PR. I've added the description that I forgot earlier :smile: 

@hemahg I will merge this now, but please take a look when you return and let me know if you spot anything that looks out of place. Thank you!

PR #1974, opened 2025-09-19
Update operator for Kafka Connect

alexcreasy:
@MikeEdgar could you take  look at this to check I've not gone completely off the rails? I'm just about to start testing it properly.

### DIFF
--- a/operator/src/main/java/com/github/streamshub/console/api/v1alpha1/spec/KafkaConnect.java
+++ b/operator/src/main/java/com/github/streamshub/console/api/v1alpha1/spec/KafkaConnect.java
@@ -1,0 +1,113 @@
+package com.github.streamshub.console.api.v1alpha1.spec;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonPropertyDescription;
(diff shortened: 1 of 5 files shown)


operator/src/main/java/com/github/streamshub/console/api/v1alpha1/spec/KafkaConnect.java
```diff
@@ -0,0 +1,113 @@
...
+    @Required
+    @JsonPropertyDescription("URL of the Kafka Connect REST API endpoint.")
+    private String url;
+
+    @JsonPropertyDescription("""
+            Indicates whether this Kafka Connect cluster is a MirrorMaker 2 deployment.
+            This affects how certain resources are displayed and managed by the console.
+            Defaults to false.
+            """)
+    private Boolean mirrorMaker = Boolean.FALSE;
```
MikeEdgar, line 41: Better to let this remain `null` and handle it (as you've already done in the ConfigurationProcessor) as possibly null. Initializing fields in the CR classes has the side effect of making changes to the CR in Kube. Sometimes that might be fine, but usually the operator should only make change to the status section itself.

```suggestion
    private Boolean mirrorMaker;
```

MikeEdgar (commented):
@alexcreasy I didn't spot anything major that looks out of place, just one comment about a CR field initializer.

Issue #2091, opened 2025-10-30
Upgrade from 0.9.0 to 0.10.0 causing 401 when accessing


sebastiangaiser:
**Describe the bug**
I upgraded from 0.9.0 to 0.10.0. Since then, I only get a 502 by our ingress controller (nginx). When having a look to the logs I get a 401 by the api container:

```
console-ui   ▲ Next.js 14.2.33                                                                                                                                                                                                                                                                                       
console-ui   - Local:        http://localhost:3000                                                                                                                                                                                                                                                                   
console-api INFO exec -a "java" java -Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager -cp "." -jar /deployments/quarkus-run.jar                                                                                                                                               
console-ui   - Network:      http://0.0.0.0:3000                                                                                                                                                                                                                                                                     
console-api INFO running in /deployments                                                                                                                                                                                                                                                                             
console-ui                                                                                                                                                                                                                                                                                                           
console-ui  ✓ Starting...                                                                                                                                                                                                                                                                                            
console-ui  ✓ Ready in 77ms                                                                                                                                                                                                                                                                                          
console-api __  ____  __  _____   ___  __ ____  ______                                                                                                                                                                                                                                                               
console-api  --/ __ \/ / / / _ | / _ \/ //_/ / / / __/                                                                                                                                                                                                                                                               
console-api  -/ /_/ / /_/ / __ |/ , _/ ,< / /_/ /\ \                                                                                                                                                                                                                                                                 
console-api --\___\_\____/_/ |_/_/|_/_/|_|\____/___/                                                                                                                                                                                                                                                                 
console-api 2025-10-30 14:16:26,582 INFO  [com.git.str.con.api.sup.fac.ConsoleConfigFactory] (main) Loading console configuration from file:/deployments/console-config.yaml                                                                                                                                         
console-api 2025-10-30 14:16:26,748 WARN  [io.fab.kub.cli.dsl.int.VersionUsageUtils] (main) The client is using resource type 'kafkas' with unstable version 'v1beta2'                                                                                                                                               
console-api 2025-10-30 14:16:27,121 WARN  [io.fab.kub.cli.dsl.int.VersionUsageUtils] (main) The client is using resource type 'kafkanodepools' with unstable version 'v1beta2'                                                                                                                                       
[...]

PR #2095, opened 2025-11-03
fix window reference error

hemahg:
fix window reference error
fixes [2091](https://github.com/streamshub/console/issues/2091)

### DIFF
--- a/ui/components/NoDataErrorState.tsx
+++ b/ui/components/NoDataErrorState.tsx
@@ -33,7 +33,7 @@ export function NoDataErrorState({ errors }: { errors: ApiError[] }) {
       <EmptyStateBody>
         <>
           {errors.map((err) => (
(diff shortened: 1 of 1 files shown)


ui/components/NoDataErrorState.tsx
MikeEdgar: Not directly related to this change, but can we add a `key` to this fragment? It logs a stack trace that could be presumed to be a bug in some cases.

```suggestion
              <React.Fragment key={err.id}>
```

The error I see is:
```
[13:32:00.381] ERROR (console/3284862): Warning: Each child in a list should have a unique "key" prop.

Check the top-level render call using <Fragment>. See https://reactjs.org/link/warning-keys for more information.
    at NoDataErrorState (webpack-internal:///(ssr)/./components/NoDataErrorState.tsx:15:29)
```

PR #2105, opened 2025-11-05
Add filters to Overview page: Filter by Broker, Time, and Topic

hemahg:
Add filters to Overview page to

- [x] filter by broker
- [x] filter by time
- [x] Filter by topic

### DIFF
--- a/ui/app/[locale]/(authorized)/kafka/[kafkaId]/overview/page.tsx
+++ b/ui/app/[locale]/(authorized)/kafka/[kafkaId]/overview/page.tsx
@@ -36,6 +36,8 @@ export default async function OverviewPage({
   const topicsForCharts = getTopics(params.kafkaId, {
     fields: "name",
     pageSize: 100,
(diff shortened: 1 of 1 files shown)


api/src/main/java/com/github/streamshub/console/api/KafkaClustersResource.java
```diff
@@ -184,11 +184,15 @@ public CompletionStage<Response> describeCluster(
                                 KafkaCluster.Fields.NODE_POOLS,
                                 KafkaCluster.Fields.CRUISE_CONTROL_ENABLED,
                             }))
-            List<String> fields) {
+            List<String> fields,
+
+            @Parameter(description = "Time range for metrics in minutes")
+            @QueryParam("duration") 
+            Integer durationMinutes) {
```
MikeEdgar: Maybe better to add the `@DefaultValue` and make the arg an `int`, then the null checking can be removed elsewhere.

PR #2109, opened 2025-11-07
Fix OIDC logout redirect to always navigate to Keycloak login page

hemahg:
Ensure consistent logout behavior for OIDC authentication.

Previously, users were redirected to the sign-in ("/") page after logout
when multiple Kafka clusters were listed. This caused the console to
redirect incorrectly instead of showing the Keycloak login screen.

Updated `handleLogout` to always redirect through the OIDC end session
endpoint, ensuring users are properly logged out from Keycloak and
returned to the login page after logout.

### DIFF
--- a/ui/utils/logout.ts
+++ b/ui/utils/logout.ts
@@ -8,29 +8,37 @@ export async function handleLogout() {
 
   const oidc = await oidcSource();
 
-  if (oidc.isEnabled()) {
-    const session = await getSession();
-    const idToken = session?.idToken;
+  try {
+    // Check if OIDC is configured
+    if (oidc.isEnabled()) {
+      const session = await getSession();
+      const idToken = session?.idToken;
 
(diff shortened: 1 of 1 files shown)


ui/utils/logout.ts
```diff
@@ -8,29 +8,37 @@ export async function handleLogout() {
...
+
+      const discovery = await fetch(`${oidc.provider!.wellKnown}`);
+      const discoveryDoc = await discovery.json();
+
+      const endSessionEndpoint = discoveryDoc.end_session_endpoint;
+
+      if (endSessionEndpoint) {
+        // Build logout URL (with or without id_token_hint)
+        const logoutUrl = `${endSessionEndpoint}${
+          idToken ? `?id_token_hint=${encodeURIComponent(idToken)}&` : "?"
```
MikeEdgar: What if the `idToken` is null/undefined here? Will it just be an empty parameter?

PR #2116, opened 2025-11-10
Update storybook packages

alexcreasy:
I've added a workaround to a bug that prevented Next.js 14 and Storybook 10.x.x working together. I've bumped the version of storybook and its dependencies. Also I've removed an unnecessary compiler warning override and handled the possible undefined exception.


MikeEdgar (approved):
👍

PR #2154, opened 2025-11-28
[GHA] Fix env variable name for systemtests workflow

jankalinic:


jankalinic:
@hemahg  this will fix env passing for the current strimzi bump PR.

PR #2166, opened 2025-12-03
[ST] Strimzi CRDs API v1

jankalinic:
Currently while running tests you may experience topic operator crashlooping (similar to https://github.com/orgs/strimzi/discussions/12173). This is because even though strimzi states parallel support for older api it does not account for kafkatopics and kafkausers - they are required to be v1 as described in (https://artifacthub.io/packages/helm/strimzi/strimzi-kafka-operator).


MikeEdgar (approved):
LGTM. I've been looking at this API change also, and I think we _can_ continue to use `v1beta2` endpoints, but we need to be careful about it. It would be good in the API server module to avoid forcing a minimum Strimzi version - I'm working on a fix for that this week and will open a PR for it.

PR #2181, opened 2025-12-10
[ST] Kafka Rebalance ST

jankalinic:
This PR adds a systemtest for a rebalance feature, deploying Kafka with CC and KafkaRebalance CRs, and using Console UI to verify data and approval of the generated proposals.

### DIFF
--- a/api/src/main/java/com/github/streamshub/console/api/TopicsResource.java
+++ b/api/src/main/java/com/github/streamshub/console/api/TopicsResource.java
@@ -41,6 +41,7 @@ import com.github.streamshub.console.api.model.ListFetchParams;
 import com.github.streamshub.console.api.model.NewTopic;
 import com.github.streamshub.console.api.model.Topic;
 import com.github.streamshub.console.api.model.TopicFilterParams;
+import com.github.streamshub.console.api.model.TopicMetrics;
 import com.github.streamshub.console.api.model.TopicPatch;
 import com.github.streamshub.console.api.security.Authorized;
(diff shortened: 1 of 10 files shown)


systemtests/src/test/java/com/github/streamshub/systemtests/kafka/RebalanceST.java
```diff
@@ -0,0 +1,99 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.util.Locale;
+import java.util.Map;
+
+import static com.github.streamshub.systemtests.utils.Utils.getTestCaseConfig;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@Tag(TestTags.REGRESSION)
+public class RebalanceST extends AbstractST {
```
MikeEdgar:
```suggestion
class RebalanceST extends AbstractST {
```

MikeEdgar:
It seems like `${{ env.ARTIFACTS_RUN_ID }}` isn't available in the job? I think the other download action had some internal logic to find it somehow. Should `Get previously built artifacts` be moved from `generate-matrix` to `run-systemtests` ?

jankalinic:
I will try this whole setup on my fork repo to avoid extra PRs, i don't think just removing unzip like that will work. The workflow action should download it to a path from which it may be necessary to move the tgz.

jankalinic:
/systemtests run --testcase=ConsumerST#testResetConsumerOffsetAllTopicsAllPartitions --retry-count=0

PR #2229, opened 2026-01-08
Expose Kafka cluster kind (Strimzi vs Kroxylicious) in API metadata

hemahg:
Expose Kafka cluster kind (Strimzi vs Kroxylicious) in API metadata

### DIFF
--- a/api/src/main/java/com/github/streamshub/console/api/model/KafkaCluster.java
+++ b/api/src/main/java/com/github/streamshub/console/api/model/KafkaCluster.java
@@ -154,10 +154,8 @@ public class KafkaCluster extends KubeApiResource<KafkaCluster.Attributes, Kafka
         Boolean reconciliationPaused;
 
         @JsonProperty
-        @Schema(description = """
-                    Kubernetes resource kind backing this Kafka cluster.
-                    implementation = ClusterKind.class
-                    """)
+        @Schema(description = "Kubernetes resource kind backing this Kafka cluster.",
+                implementation = ClusterKind.class)
         ClusterKind kind;
     }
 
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/github/streamshub/console/config/KafkaClusterConfig.java
```diff
@@ -37,6 +37,15 @@ public class KafkaClusterConfig implements Named {
...
     private String schemaRegistry;
+
+    /**
+     * Kubernetes resource kind backing this Kafka cluster.
+     *
+     * Supported values:
+     * - kafkas.kafka.strimzi.io
+     * - virtualkafkaclusters.kroxylicious.io
+     */
+    private String kind;
```
MikeEdgar: Instead of a `String` type, we can use an enum, which will make these more strictly-defined. You can also have a version of this in the operator module for the CRD.

```java
public enum ClusterKind {
    STRIMI_KAFKA("kafkas.kafka.strimzi.io"),
    KROXYLICIOUS_VIRTUAL_KAFKA_CLUSTER("virtualkafkaclusters.kroxylicious.io");

    private final String value;

    private ClusterKind(String value) {
        this.value = value;
    }

    @JsonValue
    public String getValue() {
        return value;
    }

    @JsonCreator
    public static ClusterKind forValue(String value) {
        for (var kind : values()) {
            if (kind.value.equals(value)) {
                return kind;
            }
        }
        throw new IllegalArgumentException("Unknown cluster kind: " + value);
    }
}
```

hemahg: Thanks for the feedback, Mike. The enum approach is much cleaner and improves type safety — it definitely simplified the PR. I have updated the PR accordingly

PR #2239, opened 2026-01-13
feat(ui): disable metrics and display info message for Virtual Kafka Clusters

hemahg:
feat(ui): disable metrics and display info message for Virtual Kafka Clusters

Screenshot 2026-01-14 at 5 17 10 PM

### DIFF
--- a/ui/api/kafka/schema.ts
+++ b/ui/api/kafka/schema.ts
@@ -11,6 +11,7 @@ export const ClusterListSchema = z.object({
   type: z.literal("kafkas"),
   meta: z.object({
     configured: z.boolean(),
+    kind: KafkaClusterKindSchema,
     authentication: z
       .union([
         z.object({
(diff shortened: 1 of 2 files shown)


MikeEdgar:
> @MikeEdgar I’ve added a “Virtual” label next to the Kafka cluster name and used an info icon to indicate that it’s a virtual cluster. I wasn’t sure which other icon would be more appropriate, so I’d appreciate your thoughts on this.
> 
> I’ve also disabled metrics when the cluster is a Kroxylicious virtual cluster. Please let me know if the icon choice, wording, or behavior makes sense.

This looks good, thanks @hemahg . That label would be good on the main Kafka listing page too.

hemahg:
Added the label in the main Kafka listing page too.

MikeEdgar (approved):
Code looks good! I haven't had a chance to run it yet, but I don't see any issues.

PR #2252, opened 2026-01-19
feat: add endpoint to filter metrics by topic name

hemahg:
feat: add endpoint to filter metrics by topic name


api/src/main/java/com/github/streamshub/console/api/TopicsResource.java
```diff
@@ -408,4 +409,26 @@ public CompletionStage<Response> patchTopic(
                 .thenApply(Response.ResponseBuilder::build);
     }
 
+    @Path("{topicId}/metrics")
```
MikeEdgar: I forgot about it for the node metrics endpoint in the other PR, but since we are adding a new sub-resource, we also need to allow for it to be configured in the authorization rules. The allowed types are in `com.github.streamshub.console.config.security.ResourceTypes` and we'll need them for `topics/metrics` and `nodes/metrics`.

PR #2272, opened 2026-01-27
Derive Kafka version from feature metadata for unmanaged clusters

MikeEdgar:
Without a Strimzi Kafka resource associated with a Kafka cluster, the console is currently missing version information. However, the `metadata.version` feature information available from the Admin interface provides a means to derive the version, at least the major and minor components. This PR adds that version information to the UI for non-managed clusters and provides a tooltip regarding the information source.

Cluster list
image

Cluster detail
image


hemahg (approved):
LGTM

PR #2292, opened 2026-02-05
Fix typeahead topic input not accepting manually entered topic names

hemahg:
Fix typeahead topic input not accepting manually entered topic names


jankalinic (approved):
Looks good. I did not find anything incorrect. I do not main the TS, so i can't tell 100%.

PR #2323, opened 2026-02-12
Migrate from Next.js 14 / React 18 to Next.js 15 / React 19

MikeEdgar:
- convert parameters to be promises, making server components async
- refactoring of ResponsiveTable that removes unnecessary ResponsiveTh and ResponsiveTd components. The application does not currently utilize auto-hiding columns. This can be revisited in the future if needed.
- remove `use-resize-observer` and replace with native implementation
- Fixes to theme icon selection
- Additional async function promise conversion
- Update timestamp when using in-app refresh button
- Fix cluster connection detail drawer
- Add missing project column in cluster table when OIDC is enabled
- Remove broken auto-refresh on group members tab, topic groups listing, and topic partitions listing


ui/package.json
```diff
@@ -29,33 +29,32 @@
     "@tanstack/react-virtual": "^3.13.18",
     "@types/js-yaml": "^4.0.9",
     "@types/node": "25.2.3",
-    "@types/react": "18.3.12",
-    "@types/react-dom": "18.3.1",
+    "@types/react": "19.0.6",
+    "@types/react-dom": "19.0.2",
```
hemahg: @types/react and @types/react-dom is defined in devDependencies so we can remove it from the dependencies

jankalinic (commented):
Looks great, thanks! 

I will run those tests locally as well before giving an approve. 

If the playwright assertions work reliably for every / most of the cases, we could refactor the STs to save exec time. Also thanks for the related async update for kubetest4j!

jankalinic (approved):
Thank you so much for this PR!

MikeEdgar:
@alexcreasy this needs to be merged and 0.12.1 released, but please still give your feedback on the changes in case you find anything we can fix in the next patch :pray:

PR #2332, opened 2026-02-18
[ST] Add kafka credentials login test

jankalinic:
Adds test scenario, where credentials are not provided in the console CR spec for kafka cluster. This allows entering kafka user credentials as login/password. Testcase verifies correct display of basic UI elements of kafka, nodes, topics. It also tests connection with kafka by pausing and resuming reconciliation.


jankalinic:
/systemtests run --testcase=KafkaCredentialsST --retry-count=0

PR #2418, opened 2026-03-16
[Operator] Create Route resources when supported

### LINKED ISSUE #1471: [Improvement] Allows users to leave hostname empty and let OpenShift cluster set the hostname instead
Currently the users has to specify hostname if Console CR. This is fine when user uses Ingress resources, but on OpenShift cluster route can be used without ingress and OpenShift itself will determine the hostname from resource name, namespace, and cluster domain. This could be a nice UX improvement for OpenShift users.

On the other hand, when hostname won't be required, it could cause issues when users won't specify it for Ingress resources so the proper approach should be discuss.

### LINKED ISSUE #2622: OIDC redirects need to force HTTPS behind TLS-terminated ingress
OIDC redirects need to force HTTPS behind TLS-terminated ingress. We need a configuration option in the OIDC section of the security configuration to indicate HTTPS is to be forced for redirect URLs originating from the OIDC framework.

jankalinic:
- Modify operator to create `Route` resources when supported by the cluster
- Otherwise, create Ingress resources as before. 
- The `hostname` field of the Console CR's spec is no longer required in the CRD schema. However, it is functionally required by the operator when Routes are not supported by the cluster.

This change also re-orders the reconcile process to move Ingress/Route handling after Console deployment. The deployment no longer depends on the ingress host name now that Next.js is not in use.

Closes #1471
Fixes #2622


MikeEdgar (commented):
I'll spend some time to go through this more carefully once the 0.12.1 release is out. Like you've mentioned, I think the tests in the `operator` module need some adjustments.

operator/src/main/java/com/github/streamshub/console/dependents/ConsoleIngress.java
```diff
@@ -33,6 +34,13 @@ public String resourceName() {
     @Override
     protected Ingress desired(Console primary, Context<Console> context) {
         String host = primary.getSpec().getHostname();
+
+        if (host == null || host.isBlank()) {
+            throw new ReconciliationException(
+                "spec.hostname is required when running on plain Kubernetes vanila clusters. " +
+                "Please set a hostname in your Console resource.");
```
MikeEdgar: This should probably go in `ConfigurationProcessor` so that we can set an error condition in the status along with any other validation errors that could occur. It wouldn't be throwing an exception in there, but it blocks progress on reconciliation because the CR will be considered invalid.

Also, the `@Required` will need to be removed from the `hostname` in `ConsoleSpec`.

operator/src/main/java/com/github/streamshub/console/ConsoleReconciler.java
```diff
@@ -140,17 +140,26 @@
             @Dependent(
                     name = ConsoleIngress.NAME,
                     type = ConsoleIngress.class,
+                    reconcilePrecondition = ConsoleIngress.Precondition.class,
+                    activationCondition = ConsoleIngress.Precondition.class,
```
MikeEdgar:
```suggestion
                    reconcilePrecondition = ConsoleIngress.Precondition.class,
                    activationCondition = ConsoleIngress.Precondition.class,
                    readyPostcondition = IngressOrRouteReadyCondition.class,
```

operator/src/main/java/com/github/streamshub/console/ConsoleReconciler.java
```diff
@@ -140,17 +140,26 @@
...
-                    readyPostcondition = IngressReadyCondition.class),
+                    }),
+            @Dependent(
+                    name = ConsoleRoute.NAME,
+                    type = ConsoleRoute.class,
+                    reconcilePrecondition = ConsoleRoute.Precondition.class,
+                    activationCondition = ConsoleRoute.Precondition.class,
+                    dependsOn = {
+                        ConsoleService.NAME
+                    }),
```
MikeEdgar:
```suggestion
            @Dependent(
                    name = ConsoleRoute.NAME,
                    type = ConsoleRoute.class,
                    reconcilePrecondition = ConsoleRoute.Precondition.class,
                    activationCondition = ConsoleRoute.Precondition.class,
                    readyPostcondition = IngressOrRouteReadyCondition.class,
                    dependsOn = {
                        ConsoleService.NAME
                    }),
```

MikeEdgar (changes requested):
I looked into the test failures and there are a few things that need to be adjusted to correct. The suggestions about adding `readyPostcondition` to the ingress and route is one part, and the other part is in the tests themselves.

Since the presence of the `Route` API will trigger the new logic, we need to make sure it is only present in the Kube server for the (single) test that needs it. In the `ConsoleReconcilerTestBase` class `setUp` method, add a JUnit `TestInfo testInfo` argument that you can use to conditionally create the `Route` CRD based on a tag. Something like:
```java
if (testInfo.getTags().contains("requires:routes.route.openshift.io")) {
    // create the route CRD
} else {
    // delete any route instances and the CRD
    try {
        delete(client.resources(Route.class).inAnyNamespace());
        client.resources(CustomResourceDefinition.class).withName("routes.route.openshift.io").delete();
    } catch (KubernetesClientException e) {
        // Ignore
    }
}
```

Then you can tag the one test that needs a Route like this:
```java
@Test
@Tag("requires:routes.route.openshift.io")
void testConsoleReconciliationWithOpenShiftMonitoring() {
    // the test
}
```

Future tests you add to check the new Route logic can use the same tag.

MikeEdgar (approved):

MikeEdgar:
@jankalinic I'm not sure if it will allow you to approve your own PR, but please review the current state of this and let me know if you spot any issues.

operator/src/main/java/com/github/streamshub/console/ConsoleReconciler.java
```diff
@@ -137,22 +133,27 @@
...
-                        ConsoleSecret.NAME,
-                        ConsoleIngress.NAME
+                        ConsoleSecret.NAME
                     },
                     readyPostcondition = DeploymentReadyCondition.class),
+            // Only one of ingress or route will be active for Kubernetes/OpenShift cluster.
+            @Dependent(
+                    name = ConsoleIngress.NAME,
+                    type = ConsoleIngress.class,
+                    reconcilePrecondition = ConsoleIngress.Precondition.class,
```
MikeEdgar, line 148: Keeping `reconcilePrecondition` instead of `activationCondition` here so that clusters upgrading to the version with this feature will have the old `Ingress` removed. Otherwise, we need a separate process for deleting it. Either way it's not ideal.

jankalinic:
Can't approve the PR, but go ahead with the merge when you feel like it (i don't want to merge it in case you'd like to add some last changes e.g. if something came to your mind).

PR #2485, opened 2026-04-15
Fix several concurrency issues

MikeEdgar:
- use ConcurrentHashMap in GroupService for map populated from potentially multiple threads (relates to #2475)
- improve concurrency handling of Kafka contexts to ensure in-use connections are not closed when an update to the Kafka CR replaces the connection configurations.


jankalinic (approved):

MikeEdgar:
The `testVariableConsumerGroupNames` seems a little flaky, different case failing each time. I'm going to merge this one.

PR #2525, opened 2026-04-27
Add eslint configuration, fix errors, and enable in CI

MikeEdgar:
Enables linting for the UI with eslint for JavaScript and Typescript files. This is also added to the CI job along with a concurrency check to cancel older jobs for the same PR/branch when new ones are started.

### DIFF
--- a/api/src/main/webui/eslint.config.js
+++ b/api/src/main/webui/eslint.config.js
@@ -1,0 +1,8 @@
+import js from '@eslint/js';
+import { defineConfig } from 'eslint/config';
+import tseslint from 'typescript-eslint';
+
+export default defineConfig(
+  js.configs.recommended,
+  tseslint.configs.recommended,
+);
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -45,13 +45,17 @@ updates:
       patterns:
       - "@patternfly/patternfly"
       - "@patternfly/react-*"
+    playwright:
+      patterns:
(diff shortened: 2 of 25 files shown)


jankalinic (approved):

api/src/main/webui/eslint.config.js
```diff
@@ -0,0 +1,8 @@
+import js from '@eslint/js';
+import { defineConfig } from 'eslint/config';
+import tseslint from 'typescript-eslint';
+
+export default defineConfig(
```
alexcreasy, line 5: This is a good, but sparse foundation. I would probably as a minimum add the react-hooks plugin to try to ensure we're not shooting ourselves in the foot, hooks are notoriously easy to get wrong leading to bizarre behaviour that's pretty hard to debug.

alexcreasy (commented):
Hey Mike,

This looks pretty good in general, is the intention just to give us something basic to work with? The ruleset is pretty minimal, that's not necessarily bad thing right now and we can improve it. 

I've left one comment inline about one high value addition, that said I'm happy to sign off if you want to get this in an iterate, rather than potentially creating more work immediately.

api/src/main/webui/eslint.config.js
MikeEdgar: Issue #2529 is open now to track that. We can add shortly in another PR along with any fixes necessary.

PR #2539, opened 2026-04-30
Use React data-view for Kafka clusters table

### LINKED ISSUE #2501: Client-side React UI: Migrate primary listing tables to PatternFly Data View
We should consider migrating to the Patternfly Data view for most or all of the listing pages in the Console. This appears to be a good way to standardize and eliminate some duplication throughout those pages.

https://www.patternfly.org/extensions/data-view/overview

Tables to be converted
- [x] ~Kafka cluster listing~ #2539 
- [x] Topics #2692 
- [x] Nodes
- [x] Rebalances
- [ ] Kafka users
- [ ] Kafka connectors & connect clusters
- [ ] Groups

MikeEdgar:
For https://github.com/streamshub/console/issues/2501

Subsequent changes will be coming to migrate all tables to use the new component.


jankalinic (approved):

api/src/main/webui/src/api/client.ts
```diff
@@ -68,17 +68,7 @@ class ApiClient {
       return {} as T;
     }
 
-    const data = await response.json();
-
-    if (!response.ok) {
```
alexcreasy, line 73: I'm just wondering, what's the intent with this change? Are we not risking errors going unhandled? We're not handling them here and we're also throwing away any error data - I'm thinking of the case where the backend errors out in an unexpected way and doesn't give a clean response - for the sake of example it returns a 500 so there's nothing in the errors array we have no way of knowing or reacting to that now?

api/src/main/webui/src/api/types.ts
```diff
@@ -33,12 +51,12 @@ export interface MetaWithPrivileges {
...
 
-export interface Resource<T = Record<string, unknown>> {
+export interface Resource {
   type: string;
   id: string;
-  attributes?: T;
+  attributes?: Record<string, unknown>;
   relationships?: Record<string, { data: ResourceIdentifier | ResourceIdentifier[] }>;
-  meta?: Record<string, unknown>;
+  meta?: object;
```
alexcreasy: I'm not sure about this? object is a very problematic type, it allows almost anything, objects, arrays, functions, regexes and probably more - what it essentially translates to is "any non-primitive value". 

Record<string, unknown> is actually a slightly stronger type for this if we known all the keys will be string, which I assume they will?

api/src/main/webui/src/api/types.ts
```diff
@@ -33,12 +51,12 @@ export interface MetaWithPrivileges {
   privileges?: string[];
 }
 
-export interface Resource<T = Record<string, unknown>> {
+export interface Resource {
```
alexcreasy, line 65: In losing the generic type are we sure it will always be the default type? I haven't looked deeply into where this is used in fairness, but just worried this might be a regression in type safety.

api/src/main/webui/src/components/common/ResourceListDataView.tsx
```diff
@@ -0,0 +1,494 @@
...
+    if (hasFilters || sortBy) {
+      setCurrentPage(listResponse?.meta?.page?.pageNumber ?? 1);
+    }
+  }, [filters, sortBy]);
+
+  /*
+   * synthetic URLSearchParams that includes the current page number.
+   * Uses local state to avoid flashing during navigation.
+   */
+  const paginationSearchParams = useMemo(() => {
```
alexcreasy, line 233: Just wondering why you have 2 memos that manipulate URL params and whether it might make sense condense them into 1 to avoid any potential strange issues or race conditions? (

api/src/main/webui/src/components/common/ResourceListDataView.tsx
```diff
@@ -0,0 +1,494 @@
...
+    clearAllFilters();
+    setPendingFilters({});
+  }, [clearAllFilters]);
+
+  const handleFilterChange = useCallback((_key: string, newValues: Partial<Record<string, string | string[]>>) => {
+    onSetFilters(newValues as Record<string, string | string[]>);
+  }, [onSetFilters]);
+
+  // Custom hook for text filter handlers
+  const useTextFilterHandlers = (filterId: string) => {
```
alexcreasy: This needs to be defined elsewhere, it's a strong convention in React that hooks need to be declared at the top level. You could move it outside the component declaration in this file, or move it to its own file.

api/src/main/webui/src/components/common/ResourceListDataView.tsx
```diff
@@ -0,0 +1,494 @@
...
+    const handleChange = useCallback((_event: React.FormEvent<HTMLInputElement> | undefined, value: string) => {
+      if (value) {
+        setPendingFilters(prev => ({ ...prev, [filterId]: value }));
+      } else {
+        onSetFilters({ [filterId]: '' });
+        setPendingFilters(prev => {
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
alexcreasy: We could look into changing the linter config slightly so these overrides aren't necessary. I asked Claude and it offered up this config change:

``` 
rules: {
  '@typescript-eslint/no-unused-vars': ['error', {
    argsIgnorePattern: '^_',
    ignoreRestSiblings: true,  // ← Allows destructuring pattern
  }],
},
```

This should allow a pass for variables starting with _ (or _ itself). Also it wouldn't complain about restructuring as a bonus where that's a pretty common pattern to just exclude a few items?

We can look at this separately, it would be nice to not have to include this many lint overrides though I think.

alexcreasy (changes requested):
Overall it's a really nice improvement, there's a couple of points I've raised - the one about ignoring the api error response is only potentially serious issue. The rest are just generally some small changes to promote best practice.

api/src/main/webui/src/api/types.ts
MikeEdgar: I'll push an update to define an abstract meta type that is `Record<string, unknown>`. We can initially use it here, but eventually these type interfaces can be cleaned up quite a bit and re-use it throughout.

api/src/main/webui/src/api/types.ts
MikeEdgar: This PR pushes the generic type up a level to the new `ListResponse<T extends Resource>` where the type of the `data` is `T` and each `T` will extend this `Resource` type. Only the `KafkaCluster` is extending `Resource` at this point, but it will be expanded as other list endpoints make use of the new component.

api/src/main/webui/src/components/common/ResourceListDataView.tsx
MikeEdgar: I was seeing some race conditions using a single memo to manipulate the params for both `useDataViewSort` and `useDataViewPagination`. I couldn't really narrow down to why or where it was happening, but I am definitely open to doing this with less code/functions if we can.

api/src/main/webui/src/components/common/ResourceListDataView.tsx
MikeEdgar: Applied your suggestion to eslint configuration

api/src/main/webui/src/components/common/ResourceListDataView.tsx
MikeEdgar: I moved this hook at the top level outside of the component. Is it mainly just a convention or are there technical reasons for doing so? I'm all for sticking to conventions, just curious.

api/src/main/webui/src/api/client.ts
MikeEdgar: I added back the error handling in client.ts and also updated the component to accept the tanstack response object (which includes the data, loading status, and error), so I _think_ that should cover it more completely. I also removed the top-level error markup since it is passed fully down into the component now.

MikeEdgar:
@alexcreasy I believe I've addressed your review comments, but let me know if something is still missing or out of place. The PR branch has been rebased, but the original commit is unchanged and your feedback is addressed in the second commit. Thanks!

api/src/main/webui/src/components/common/ResourceListDataView.tsx
alexcreasy: The biggest technical issue I can think of that you might run into is that by defining a function inside another function in JavaScript you've created a closure with its own copy of the surrounding data, including the props the outer function was called with. This means you could in some circumstances end up with your nested hook using stale data from a previous render, leading to subtle bugs that are very difficult to figure out.

Other than that I think the convention is probably so that developers don't mix up hooks with helper functions or event handlers and end up forgetting the "rules of hooks", like not calling them in conditionals or event handlers etc.

api/src/main/webui/src/components/common/ResourceListDataView.tsx
alexcreasy: No worries, if it fixed some issues then I'm good.

alexcreasy (approved):

MikeEdgar:
After the update to PF 6.5 on `main` since the last rebase, this now appears to be blocked on https://github.com/patternfly/react-data-view/issues/662.

PR #2546, opened 2026-05-05
Fixes and Enhancements to group reset functionality

MikeEdgar:
Refine the group offset reset functionality in new client-side React UI.

- simplify topic/partition selection
- simplify date and time zone selection
- display current + new offsets in dry-run results

Assisted-by: IBM Bob


jankalinic (approved):
LGTM, can't run test's against this right now, I'll fix the login page shortly, if you want to wait for the reset offset ST.

api/src/main/webui/src/utils/cliCommandGenerator.ts
alexcreasy, line 15: It looks deliberate that the bootstrap server is a literal string rather than a templatevar, but I just wanted to check this is the case? 

It's not actually changed in this PR but it just stuck out at me as something to check!

alexcreasy (approved):
1 comment I wasn't 100% sure on. 

I tested this on our cloud deployment and it works using dry runs. No complaints with the code. UX seems straightforward for doing this.

LGTM.

api/src/main/webui/src/utils/cliCommandGenerator.ts
MikeEdgar: That's right - we keep that as a variable because there may be multiple bootstrap addresses for them to choose from.

PR #2547, opened 2026-05-05
[0.12.x] Update several dependencies (NPM and Maven)

MikeEdgar:
The same updates are applied to `main` in #2549.


MikeEdgar:
@jankalinic wdyt? Flaky tests OK?

alexcreasy (approved):
All fine, if the test failures are just flakes seems fine to proceed whilst we're in transition.

PR #2605, opened 2026-05-21
Form-based Kafka SCRAM and OAUTHBEARER authentication

### LINKED ISSUE #2496: Client-side React UI: Add per-cluster Kafka authentication
Add per-cluster login functionality to the API server, with in-memory credential encryption on the server and client session is maintained using a `SameSite=Strict`, `Secure`, and `HttpOnly` session cookie.

MikeEdgar:
This re-implements the form-based Kafka authentication from the previous Next.js UI. Rather than a separate login page, this PR introduces a login modal that will be used whenever the back-end responds with a 401/unauthorized. Session cookies are scoped to each Kafka cluster the user has logged in to.

Closes #2496 

- Updated documentation with deprecation and security notices for form-based authentication.
- Restored "Authentication" column in clusters table 
  image
- Modal appears when accessing cluster where user is unauthenticated
  image
- Toggling between clusters does not prompt to log-in while session is active.

### DIFF
--- a/api/src/main/java/com/github/streamshub/console/api/security/kafka/KafkaIdentityProviderLogin.java
+++ b/api/src/main/java/com/github/streamshub/console/api/security/kafka/KafkaIdentityProviderLogin.java
@@ -67,8 +67,10 @@ public class KafkaIdentityProviderLogin implements IdentityProvider<UsernamePass
             SaslJaasConfigCredential credential) {
 
         var promise = new CompletableFuture<SecurityIdentity>();
-        var admin = clientFactory.createAdmin(filter, adminBuilder, kafkaContext, credential);
+        @SuppressWarnings("java:S2095") // Admin must remain open until async operation completes
+        Admin admin = clientFactory.createAdmin(filter, adminBuilder, kafkaContext, credential);
 
+        // Close admin after async operation completes (success or failure)
         promise.whenComplete((identity, error) -> {
             try {
                 admin.close();
--- a/api/src/main/java/com/github/streamshub/console/api/security/kafka/KafkaIdentityProviderLogin.java
+++ b/api/src/main/java/com/github/streamshub/console/api/security/kafka/KafkaIdentityProviderLogin.java
@@ -10,6 +10,7 @@ import jakarta.inject.Inject;
 import jakarta.json.Json;
 
 import org.apache.kafka.clients.admin.Admin;
+import org.jboss.logging.Logger;
 
 import com.github.streamshub.console.api.ClientFactory;
 import com.github.streamshub.console.api.security.IdentitySupport;
@@ -27,6 +28,9 @@ import io.smallrye.mutiny.Uni;
 
 @ApplicationScoped
 public class KafkaIdentityProviderLogin implements IdentityProvider<UsernamePasswordAuthenticationRequest> {
+
+    @Inject
+    Logger logger;
 
     @Inject
     IdentitySupport identities;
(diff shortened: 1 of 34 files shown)


MikeEdgar:
Once #2611 is merged, this will be rebased and (presumably) have some additional changes due to the lint configuration changes there.

MikeEdgar:
This PR will also need updates to the documentation to advise that form-based authentication is not advised in security-critical environments (i.e. production) and that OIDC should be used instead.

MikeEdgar:
@AratiAjit it won't allow me to request a review from you, so I'll try to just mention you instead :)

api/src/main/java/com/github/streamshub/console/api/security/kafka/KafkaIdentityProviderLogin.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    private Uni<SecurityIdentity> authenticate(
+            KafkaContext kafkaContext,
+            String principalName,
+            SaslJaasConfigCredential credential) {
+
+        var promise = new CompletableFuture<SecurityIdentity>();
+
+        try (var admin = clientFactory.createAdmin(filter, adminBuilder, kafkaContext, credential)) {
```
alexcreasy: Claude threw this out there, I'm not familiar with the admin client, so I offer this up in case it's helpful :) 

"
 1. Admin client closed before async operation completes (KafkaIdentityProviderLogin.java:975)

  try (var admin = clientFactory.createAdmin(filter, adminBuilder, kafkaContext, credential)) {
      admin.describeCluster()
          .clusterId()
          .toCompletionStage()
          .thenApply(...)
          .exceptionally(promise::completeExceptionally);
  }  // ← admin.close() fires HERE, before the CompletionStage resolves

  The try-with-resources closes the Admin client immediately after firing the async describeCluster() call, not when it completes. This is a race condition — the admin client may be closed while the Kafka request is in flight, causing authentication to silently fail or throw. The admin close should be chained into
  the completion stage (e.g., in a whenComplete handler).
"

jankalinic (approved):

AratiAjit:
The doc update about the deprecation warning looks good to me.

api/src/main/webui/src/components/auth/KafkaAuthProvider.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+  const handleAuthError = useCallback((error: unknown, clusterId?: string): boolean => {
+    // Check if this is a 401 error
+    if (error instanceof ApiError && error.status === 401) {
+      const targetClusterId = clusterId || kafkaId;
+      
+      if (targetClusterId) {
+        // Get cluster name from the error context or use clusterId as fallback
+        const clusterName = targetClusterId;
+        showLoginModal(targetClusterId, clusterName, window.location.pathname);
```
alexcreasy: The showLoginModal function takes 4 params, 2 of which are optional - you're passing the url path as the `authMethod` param - I'm guessing it's supposed to be the `pendingNavigation` param?

Should this be: `showLoginModal(targetClusterId, clusterName, undefined, window.location.pathname);` ?

alexcreasy (changes requested):
Really good refactor - security principles are all sound IMO.

I found one small frontend issue that could break post-login redirect. Claude found a potential race condition in the backend that I think could cause arbitrary authentication failures, if it's correct.

api/src/main/java/com/github/streamshub/console/api/security/kafka/KafkaIdentityProviderLogin.java
MikeEdgar: This has been fixed - the Admin client is now closed in a `whenComplete` handler as suggested :+1:

api/src/main/webui/src/components/auth/KafkaAuthProvider.tsx
MikeEdgar: Good catch, this has been fixed as suggested.

alexcreasy (approved):

PR #2611, opened 2026-05-27
Enable eslint-plugin-react-hooks, fix errors

### LINKED ISSUE #2529: Client-side React UI: Add react-hooks eslint plugin
Add/enable the [react-hooks](https://react.dev/reference/eslint-plugin-react-hooks) eslint plugin and make any fixes necessary. This is a followup to PR #2525

MikeEdgar:
Closes #2529

### DIFF
--- a/api/src/main/webui/src/components/common/ResourceListDataView.tsx
+++ b/api/src/main/webui/src/components/common/ResourceListDataView.tsx
@@ -12,6 +12,7 @@ import {
   DataViewTr,
   DataViewState,
   useDataViewSort,
+  DataViewTextFilterProps,
 } from '@patternfly/react-data-view';
 /*
  * The following import is a work-around for
@@ -45,60 +46,67 @@ const perPageOptions = [
 
 const DEFAULT_PAGE_SIZE = 10;
 
-// Component wrapper for text filter to avoid hooks-in-callback issue
-const TextFilterWrapper: React.FC<{
-  name: string;
-  filter: {
-    type: 'text' | 'checkbox';
-    title: string;
-    placeholder: string;
-    initialValue?: string | string[];
-  };
-  pendingFilters: Record<string, string>;
-  setPendingFilters: React.Dispatch<React.SetStateAction<Record<string, string>>>;
-  onSetFilters: (newFilters: Partial<Record<string, string | string[]>>) => void;
-}> = ({ name, filter, pendingFilters, setPendingFilters, onSetFilters }) => {
-  const handleChange = (_event: React.FormEvent<HTMLInputElement> | undefined, value: string) => {
+// Custom hook for text filter handlers
+const useTextFilterHandlers = (
+  filterId: string,
+  pendingFilters: Record<string, string>,
+  setPendingFilters: React.Dispatch<React.SetStateAction<Record<string, string>>>,
+  onSetFilters: (newFilters: Partial<Record<string, string | string[]>>) => void,
+) => {
+  const handleChange = useCallback((_event: React.FormEvent<HTMLInputElement> | undefined, value: string) => {
     if (value) {
-      setPendingFilters(prev => ({ ...prev, [name]: value }));
+      setPendingFilters(prev => ({ ...prev, [filterId]: value }));
     } else {
-      onSetFilters({ [name]: '' });
+      onSetFilters({ [filterId]: '' });
       setPendingFilters(prev => {
-        const { [name]: _, ...rest } = prev;
+        const { [filterId]: _, ...rest } = prev;
         return rest;
       });
     }
-  };
+  }, [filterId, onSetFilters, setPendingFilters]);
 
-  const handleClear = () => {
-    onSetFilters({ [name]: '' });
+  const handleClear = useCallback(() => {
+    onSetFilters({ [filterId]: '' });
     setPendingFilters(prev => {
-      const { [name]: _, ...rest } = prev;
+      const { [filterId]: _, ...rest } = prev;
       return rest;
     });
-  };
+  }, [filterId, onSetFilters, setPendingFilters]);
 
-  const handleSearch = (_event: React.SyntheticEvent<HTMLButtonElement>, value: string) => {
-    const filterValue = pendingFilters[name] ?? value;
-    onSetFilters({ [name]: filterValue });
+  const handleSearch = useCallback((_event: React.SyntheticEvent<HTMLButtonElement>, value: string) => {
+    onSetFilters({ [filterId]: pendingFilters[filterId] ?? value });
     setPendingFilters(prev => {
-      const { [name]: _, ...rest } = prev;
+      const { [filterId]: _, ...rest } = prev;
       return rest;
     });
-  };
+  }, [filterId, pendingFilters, onSetFilters, setPendingFilters]);
 
-  return (
-    <DataViewTextFilter
-      filterId={name}
-      title={filter.title}
-      placeholder={filter.placeholder}
+  return { handleChange, handleClear, handleSearch };
+};
+
+type TextFilterWrapperProps = DataViewTextFilterProps & {
+  pendingFilters: Record<string, string>;
+  setPendingFilters: React.Dispatch<React.SetStateAction<Record<string, string>>>;
+  onSetFilters: (newFilters: Partial<Record<string, string | string[]>>) => void;
+};
+
+// Component wrapper for text filter to avoid hooks-in-callback issue
+const TextFilterWrapper: React.FC<TextFilterWrapperProps> = (props) => {
+  const { pendingFilters, setPendingFilters, onSetFilters, ...filterProps } = props;
+  const { handleChange, handleClear, handleSearch } = useTextFilterHandlers(
+    filterProps.filterId,
+    pendingFilters,
+    setPendingFilters,
+    onSetFilters,
+  );
+  return useMemo(() => <DataViewTextFilter
+      {...filterProps}
       onChange={handleChange}
       onClear={handleClear}
       onSearch={handleSearch}
-    />
-  );
-};
-
+    />,
+    [filterProps, handleChange, handleClear, handleSearch]);
+}
 export interface ResourceListDataViewColumnMapper {
   (
     sortBy?: string, 
@@ -196,24 +204,25 @@ export function ResourceListDataView<T extends Resource>({
    * synthetic URLSearchParams to convert single sort parameter
    * to two parameters expected by the hook
    */
+  const sortParam = useMemo(() => searchParams.get('sort'), [searchParams]);
   const sortSearchParams = useMemo(() => {
     const params = new URLSearchParams();
-    const sort = searchParams.get("sort");
-    if (sort) {
+
+    if (sortParam) {
       let sortBy: string;
 
-      if (sort.startsWith("-")) {
+      if (sortParam.startsWith("-")) {
         params.set("direction", "desc");
-        sortBy = sort.substring(1);
+        sortBy = sortParam.substring(1);
       } else {
         params.set("direction", "asc");
-        sortBy = sort;
+        sortBy = sortParam;
       }
 
       params.set("sortBy", sortBy);
     }
     return params;
-  }, [searchParams]);
+  }, [sortParam]);
 
   const { sortBy, direction, onSort } = useDataViewSort({
     searchParams: sortSearchParams,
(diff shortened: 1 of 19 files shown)


jankalinic (approved):

api/src/main/webui/src/components/common/ResourceListDataView.tsx
```diff
@@ -211,8 +227,9 @@ export function ResourceListDataView<T extends Resource>({
 
   // Update current page when API response arrives
   useEffect(() => {
-    if (listResponse?.meta?.page?.pageNumber) {
-      setCurrentPage(listResponse.meta.page.pageNumber);
+    const pageNumber = listResponse?.meta?.page?.pageNumber;
+    if (pageNumber) {
+      setTimeout(() => setCurrentPage(pageNumber), 0);
```
alexcreasy: What is the timeout for? This is triggering my spidey sense a bit because deferring the call to the next tick can be unpredictable and add unnecessary extra rendering cycles.

api/src/main/webui/src/components/common/ResourceListDataView.tsx
```diff
@@ -222,26 +239,26 @@ export function ResourceListDataView<T extends Resource>({
       Array.isArray(v) ? v.length > 0 : v?.trim().length > 0
     );
     if (hasFilters || sortBy) {
-      setCurrentPage(listResponse?.meta?.page?.pageNumber ?? 1);
+      setTimeout(() => setCurrentPage(listResponse?.meta?.page?.pageNumber ?? 1), 0);
```
alexcreasy: Same here, are they all for the same reason?

api/src/main/webui/src/components/kafka/overview/TopicChartsCard.tsx
```diff
@@ -74,7 +75,8 @@ export function TopicChartsCard({
     );
 
     if (!existsInFiltered) {
-      setSelectedTopic(null);
+      // Use setTimeout to defer state update to avoid cascading renders
+      setTimeout(() => setSelectedTopic(null), 0);
```
alexcreasy: OK presumably these are all to avoid cascading renders? I'll take a deeper look and try to suggest a better way of doing it.

api/src/main/webui/src/components/kafka/topics/AdvancedSearch.tsx
```diff
@@ -173,8 +173,10 @@ export function AdvancedSearch({
     setIsExpanded(false);
   };
 
+  // Sync search input value with URL params
   useEffect(() => {
-    setSearchInputValue(getSearchInputValue());
+    // Use setTimeout to defer state update to avoid cascading renders
+    setTimeout(() => setSearchInputValue(getSearchInputValue()), 0);
```
alexcreasy: and here.

api/src/main/webui/src/pages/kafka/groups/detail/GroupConfigurationTab.tsx
```diff
@@ -108,9 +108,10 @@ export function GroupConfigurationTab() {
   }, [allData]);
 
   // Initialize selected data sources to all sources
-  useMemo(() => {
+  useEffect(() => {
     if (dataSources.length > 0 && selectedDataSources.length === 0) {
-      setSelectedDataSources(dataSources);
+      // Use setTimeout to defer state update to avoid cascading renders
+      setTimeout(() => setSelectedDataSources(dataSources), 0);
```
alexcreasy: and here.

api/src/main/webui/src/pages/kafka/nodes/detail/NodeConfigurationTab.tsx
```diff
@@ -111,9 +111,10 @@ export function NodeConfigurationTab() {
   }, [typedAllData]);
 
   // Initialize selected data sources to all sources
-  useMemo(() => {
+  useEffect(() => {
     if (dataSources.length > 0 && selectedDataSources.length === 0) {
-      setSelectedDataSources(dataSources);
+      // Use setTimeout to defer state update to avoid cascading renders
+      setTimeout(() => setSelectedDataSources(dataSources), 0);
```
alexcreasy: here

api/src/main/webui/src/pages/kafka/topics/detail/TopicConfigurationTab.tsx
```diff
@@ -132,9 +132,10 @@ export function TopicConfigurationTab() {
   }, [typedAllData]);
 
   // Initialize selected data sources to all sources
-  useMemo(() => {
+  useEffect(() => {
     if (dataSources.length > 0 && selectedDataSources.length === 0) {
-      setSelectedDataSources(dataSources);
+      // Use setTimeout to defer state update to avoid cascading renders
+      setTimeout(() => setSelectedDataSources(dataSources), 0);
```
alexcreasy: here

api/src/main/webui/src/pages/kafka/topics/detail/TopicMessagesTab.tsx
```diff
@@ -111,10 +112,11 @@ export function TopicMessagesTab() {
         m => m.attributes.partition === part && m.attributes.offset === off
       );
       if (msg) {
-        setSelectedMessage(msg);
+        // Use setTimeout to defer state update to avoid cascading renders
+        setTimeout(() => setSelectedMessage(msg), 0);
```
alexcreasy: here

api/src/main/webui/src/pages/kafka/topics/detail/TopicMessagesTab.tsx
```diff
@@ -111,10 +112,11 @@ export function TopicMessagesTab() {
...
       }
     } else if (!selectedMessageId) {
-      setSelectedMessage(undefined);
+      setTimeout(() => setSelectedMessage(undefined), 0);
```
alexcreasy: here (I'm just adding comments now so i remember where they are)

api/src/main/webui/src/pages/kafka/topics/detail/TopicMessagesTab.tsx
```diff
@@ -174,7 +176,8 @@ export function TopicMessagesTab() {
       try {
         const parsed = JSON.parse(saved);
         if (Array.isArray(parsed)) {
-          setChosenColumns(parsed);
+          // Use setTimeout to defer state update to avoid cascading renders
+          setTimeout(() => setChosenColumns(parsed), 0);
```
alexcreasy: here

api/src/main/webui/src/components/common/ResourceListDataView.tsx
```diff
@@ -197,7 +213,7 @@ export function ResourceListDataView<T extends Resource>({
       params.set("sortBy", sortBy);
     }
     return params;
-  }, [searchParams.get("sort")]);
+  }, [searchParams]);
```
alexcreasy: Does this change fix something? It seems like it would be better to only re-render on a change in the sort param rather than the whole list?

api/src/main/webui/src/components/common/ResourceListDataView.tsx
```diff
@@ -257,7 +274,7 @@ export function ResourceListDataView<T extends Resource>({
     setBeforeCursor(undefined);
     setAfterCursor(undefined);
     pagination.onPerPageSelect(event, newPerPage);
-  }, [pagination.onPerPageSelect]);
+  }, [pagination]);
```
alexcreasy, line 266: I don't think pagination is a stable reference for the dependency array here, on every render pagination gets defined again, so the callback gets recomputed every render, so it isn't memoized.

api/src/main/webui/src/components/common/ResourceListDataView.tsx
```diff
@@ -339,12 +356,12 @@ export function ResourceListDataView<T extends Resource>({
...
       return newParams;
     }, { replace: true }); // Replace current history entry instead of pushing new one
-  }, [filters, sortBy, direction, perPage, beforeCursor, afterCursor]);
+  }, [filters, sortBy, direction, perPage, beforeCursor, afterCursor, setSearchParams]);
 
   // Define columns
   const columns: DataViewTh[] = useMemo(() => {
     return columnProvider.callback(sortBy, direction, onSort);
-  }, [sortBy, direction, ...columnProvider.dependencies]);
+  }, [sortBy, direction, onSort, columnProvider]);
```
alexcreasy, line 353: is columnProvider a stable reference to use here? It comes as a prop to this component so I'm not sure, if it is it's all good, but otherwise you might be better off depending on some of its properties directly.

api/src/main/webui/src/pages/kafka/topics/detail/TopicPartitionsTab.tsx
```diff
@@ -111,7 +111,7 @@ export function TopicPartitionsTab() {
   }, [refetch]);
 
   const topic = data?.data;
-  const partitions = topic?.attributes.partitions || [];
+  const partitions = useMemo(() => topic?.attributes.partitions || [], [topic?.attributes.partitions]);
```
alexcreasy, line 114: Why is this a memo now? It doesn't look like you're actually computing anything to memoize, so it's redundant?

api/src/main/webui/src/components/common/ResourceListDataView.tsx
```diff
@@ -45,42 +45,58 @@ const perPageOptions = [
...
-// Custom hook for text filter handlers
-const useTextFilterHandlers = (
-  filterId: string,
-  pendingFilters: Record<string, string>,
-  setPendingFilters: React.Dispatch<React.SetStateAction<Record<string, string>>>,
-  onSetFilters: (newFilters: Partial<Record<string, string | string[]>>) => void,
-) => {
-  const handleChange = useCallback((_event: React.FormEvent<HTMLInputElement> | undefined, value: string) => {
+// Component wrapper for text filter to avoid hooks-in-callback issue
+const TextFilterWrapper: React.FC<{
```
alexcreasy: Maybe use React.memo to wrap this component?  It's going to get re-rendered every time it's parent does whether its props have changed or not, which might happen quite frequently given it's part of a data table?

alexcreasy (changes requested):
There's a lot of noise in my review, sorry I've marked every instance where you've called setTimeout(..., 0) to get round cascading renders. I'll take a look through those calls tomorrow morning and come up with some actual suggestions to replace them rather than just tossing grenades at you! :)

That's the only major issue I see, I've also pointed out a few issues potentially around unstable references for hook dependencies and there's a suggestion to memoize a component to save some performance!

Other than that it's all good, I'll get you those suggestions to replace the setTimeouts ASAP.

api/src/main/webui/src/components/common/ResourceListDataView.tsx
MikeEdgar: These are the messages associated with the original.
```
  216:6  warning  React Hook useMemo has a missing dependency: 'searchParams'. Either include it or remove the dependency array                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             react-hooks/exhaustive-deps
  216:7  error    Error: Expected the dependency list to be an array of simple expressions (e.g. `x`, `x.y.z`, `x?.y?.z`)

Expected the dependency list to be an array of simple expressions (e.g. `x`, `x.y.z`, `x?.y?.z`).
```

api/src/main/webui/src/components/kafka/overview/TopicChartsCard.tsx
MikeEdgar: Yes, exactly that.
```
error  Error: Calling setState synchronously within an effect can trigger cascading renders
```

api/src/main/webui/src/components/common/ResourceListDataView.tsx
MikeEdgar: This is something I fundamentally have trouble understanding. `pagination` is assigned to the output the `useDataViewPagination` hook, much like many other `const` fields. For a given set of arguments to `useDataViewPagination`, wouldn't the result be stable? Or is it unstable because that hook returns an object literal rather than some memoized object?

MikeEdgar: I tried wrapping the call to `useDataViewPagination` with `useMemo` and this error appears:
```
Compilation Skipped: Existing memoization could not be preserved

React Compiler has skipped optimizing this component because the existing manual memoization could not be preserved. This value was memoized in source but not in compilation output.
```

Does that mean hooks are automatically memoized? Does that change your thinking about `pagination` being stable?

api/src/main/webui/src/pages/kafka/topics/detail/TopicPartitionsTab.tsx
MikeEdgar: Per eslint:

> The 'partitions' logical expression could make the dependencies of useMemo Hook (at line 152) change on every render. To fix this, wrap the initialization of 'partitions' in its own useMemo() Hook.

api/src/main/webui/src/components/common/ResourceListDataView.tsx
MikeEdgar: Removed the `setTimeout`s throughout this PR.

api/src/main/webui/src/components/common/ResourceListDataView.tsx
MikeEdgar: I think I've addressed this. Are components re-rendered without changes to their inputs?

alexcreasy (approved):
Nice, thanks for the fixes!

LGTM.

PR #2677, opened 2026-06-23
Adjust console startup probe for CPU-constrained scenarios

### LINKED ISSUE #2676: Adjust startup probe to account for scenarios with restrained CPU limits
The current startup probes only allow for 3 failures, which may result in crash looping where CPU is constrained. We should increase the failure limit and reduce the initial delay and period to 1) support faster recognition that the container has started up and 2) be more tolerant of constrained CPU scenarios.

For example, this probe will check sooner (after 3s), more frequently (every 5s), for longer (up to 120s/2m) for the container to be started.

```yaml
startupProbe:
  httpGet:
    path: /q/health/started
    port: 8080
    scheme: HTTP
  initialDelaySeconds: 3
  timeoutSeconds: 10
  periodSeconds: 5
  successThreshold: 1
  failureThreshold: 24
```

GabrielBBaldez:
Adjusts the `console-api` startup probe per #2676 so it tolerates slow startups under constrained CPU:

- `failureThreshold`: 3 -> 24
- `initialDelaySeconds`: 5 -> 3
- `periodSeconds`: 10 -> 5

The startup window becomes ~2 minutes (3 + 5*24 = 123s) instead of failing after only 3 checks and crash looping.

I kept the existing `path: /health/started` and `port: http` (rather than the `/q/health/started` and `8080` shown in the issue example): the API sets `quarkus.http.non-application-root-path` to the root path, so the health endpoints are served under `/health`, not `/q/health`. Only the timing values were changed.

Closes #2676

### DIFF
--- a/operator/src/main/resources/application.properties
+++ b/operator/src/main/resources/application.properties
@@ -17,6 +17,11 @@ quarkus.kubernetes.name=streamshub-console-operator
 quarkus.kubernetes.rbac.service-accounts.streamshub-console-operator.namespace=
 quarkus.kubernetes.env.fields."CONSOLE_DEPLOYMENT_DEFAULT_IMAGE_TAG"=metadata.labels['app.kubernetes.io/version']
 
+# Tolerate slow startup under constrained CPU, avoiding crash loops (#2676)
+quarkus.kubernetes.startup-probe.initial-delay=3s
+quarkus.kubernetes.startup-probe.period=5s
+quarkus.kubernetes.startup-probe.failure-threshold=24
+
 # Not needed. Disable to support read-only FS
 # See https://github.com/eclipse-vertx/vert.x/blob/1a28a6f0913802951dc5e6536dc99963f80b466a/vertx-core/src/main/asciidoc/filesystem.adoc?plain=1#L130
 quarkus.vertx.caching=false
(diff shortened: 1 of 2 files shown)


MikeEdgar (changes requested):
Thanks for the PR @GabrielBBaldez . Since they are currently being generated by the operator SDK, we will need to add an explicit startup probe to the operator YAML template [1] so it is used instead of the generated one. You can use the same delay, period, and failure threshold, but note that the operator's startup health endpoint is `/q/health/started`

[1] https://github.com/streamshub/console/blob/044d4cf38a7f45f518bf1134ef159fbf3096c77c/operator/src/main/kubernetes/kubernetes.yml#L343

GabrielBBaldez:
I tried the explicit probe in the operator `kubernetes.yml`, but the operator-sdk still generated the default over it (the build kept `failureThreshold: 3`). Setting `quarkus.kubernetes.startup-probe.*` in `application.properties` produces the right values in the generated manifest (`failureThreshold: 24`, `initialDelay: 3s`, `period: 5s`, path `/q/health/started`). Happy to move it back to the YAML template if there's a way to make that override stick.

MikeEdgar (approved):
Looks good, thank you for the PR @GabrielBBaldez

PR #2683, opened 2026-06-24
fix: restore Kroxy Virtual labels, Kafka version fetch error handling

MikeEdgar:
Restore the `Virtual` labels for the non-Next.js UI. Also adds a timeout and error handling (log & continue) when the Kafka version cannot be retrieved.

image

### DIFF
--- a/api/src/main/java/com/github/streamshub/console/api/service/KafkaClusterService.java
+++ b/api/src/main/java/com/github/streamshub/console/api/service/KafkaClusterService.java
@@ -72,6 +72,13 @@ public class KafkaClusterService {
     private static final String AUTHN_KEY = "authentication";
     private static final String AUTHN_METHOD_KEY = "method";
 
+    /**
+     * Options for the describeFeatures operation used to obtain a Kafka cluster
+     * version when a Strimzi Kafka CR is not present. We keep the timeout low to
+     * limit impact to list cluster operations.
+     */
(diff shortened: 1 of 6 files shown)


jankalinic (approved):

api/src/main/java/com/github/streamshub/console/api/service/KafkaClusterService.java
```diff
@@ -271,9 +272,17 @@ KafkaCluster addKafkaContextData(KafkaCluster cluster, KafkaContext kafkaContext
              * application-wide authentication.
              */
             Optional.ofNullable(kafkaContext.admin())
-                    .map(Admin::describeFeatures)
+                    .map(admin -> admin.describeFeatures(new DescribeFeaturesOptions()
+                        .timeoutMs(5000)))
```
alexcreasy: Pedantic nit: magic number use, would be better as a constant.

alexcreasy (approved):
LGTM, one pedantic nit, but approved anyway!

api/src/main/java/com/github/streamshub/console/api/service/KafkaClusterService.java
MikeEdgar: Fixed, the entire `DescribeFeaturesOptions` object with timeout option is now a constant with a comment on its use.

PR #2692, opened 2026-06-25
Migrate topic listings to use common fetch and data view component

### LINKED ISSUE #2718: [UI] Missing search button - enter
**Describe the bug**
Currently search input for Topics page filter-by name/topicId has only search input without the "search" or "enter" button next to it, in my opinion it's nicer to have that, instead of pressing enter button on a keyboard to filter. The search itself is working, this issue is just about the lack of the button.

Possibly in case when the search is dynamic and reacts to what user types at the moment, i could see removal of the search button useful, but that might be non-optimal and take resources/time.

Image

MikeEdgar:
Migrate the topics table to use the new data view component. 

This PR also reverses the toggle control to show internal topics. Previously, the toggle would show them when off and hide them when on - with label "Hide internal topics" and "on" by default. With the change, the label reads "Show internal topics" and is disabled/off by default. See screenshot below.

image

### DIFF
--- a/api/src/main/webui/src/api/hooks/useResourceList.ts
+++ b/api/src/main/webui/src/api/hooks/useResourceList.ts
@@ -16,6 +16,11 @@ export interface ResourceListPageParams {
   } | string;
 }
 
+export interface ResourceListFilterValue {
+  operator: 'eq' | 'in' | 'like';
+  value: string;
+}
+
 export interface ResourceListParams {
   /**
    * Parameters for pagination (page size, sorting, etc.)
@@ -24,7 +29,7 @@ export interface ResourceListParams {
   /**
    * Parameters for filtering by specific fields (search, etc.)
    */
-  filters?: Record<string, string | string[]>;
+  filters?: Record<string, string | string[] | ResourceListFilterValue>;
   /**
    * Comma-separated list of fields to include in the response.
    * @example
@@ -84,10 +89,21 @@ export function useResourceList<T extends Resource>(
         updatePageParams(params.page, searchParams);
       }
 
-      // Handle name filter
       if (params?.filters) {
         Object.entries(params.filters).forEach(([key, value]) => {
-          searchParams.set(`filter[${key}]`, `like,*${value}*`);
+          if (Array.isArray(value)) {
+            if (value.length > 0) {
+              searchParams.set(`filter[${key}]`, `in,${value.join(',')}`);
+            }
+            return;
+          }
+
+          if (typeof value === 'string') {
+            searchParams.set(`filter[${key}]`, `like,*${value}*`);
+            return;
+          }
+
+          searchParams.set(`filter[${key}]`, `${value.operator},${value.value}`);
         });
       }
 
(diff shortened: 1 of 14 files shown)


jankalinic (approved):

api/src/main/webui/src/api/hooks/useResourceList.ts
alexcreasy, line 82: This uses the same cache for all kafka clusters, could that lead to collisions? (genuine question, I'm not 100% sure).

Is there anything in the query params that references the cluster on the backend? If so using the same cache is perfectly safe as you won't get collisions -- if not there's a risk that you could get a cache collision and maybe `queryKey[0]` should contain a reference to the cluster?

Again, there may not be an issue here,  I just wanted to highlight this in case, finding a bug caused by this later would be a nightmare.

api/src/main/webui/src/components/common/ResourceListDataView.tsx
```diff
@@ -464,34 +557,69 @@ export function ResourceListDataView<T extends Resource>({
...
+                    />;
+                  }
+                }).filter(entry => entry !== undefined)}
+              </DataViewFilters>
+              {toggleFilters.map(([name, filter]) => (
+                <ToggleFilter
+                  key={`filter-${name}`}
+                  filterId={name}
+                  filter={filter as ResourceListToggleFilterConfig}
+                  isChecked={filters[name] === true}
```
alexcreasy: From the type declaration filters could also be strings? I think this is populated from searchParams.get - which definitely returns strings.

Just changing this to: `isChecked={String(filters[name]) === 'true'} will handle all cases and ensure we don't end up with the box being checked not matching whether hidden items are actually included.

api/src/main/webui/src/components/kafka/topics/TopicsDataView.tsx
```diff
@@ -0,0 +1,257 @@
...
+            <>
+              {t('topics.showInternalTopics')}{' '}
+              <Tooltip content={t('topics.showInternalTopicsTooltip')}>
+                <HelpIcon />
+              </Tooltip>
+            </>
+          ),
+        },
+      }}
+      columnProvider={{
```
alexcreasy: I don't know how big a deal this is, but declaring this as a literal breaks the memo in ResourceListDataView.tsx:433-435 - because columnProvider is a dependency in that memo, and since this is a literal a new object is created on each render, changing its reference value and defeating the memo clause.

To keep the memo functionality you just need to memoize columnProvider here.

Claude found this one BTW (I'd have missed it, no question!)

alexcreasy (changes requested):
Looks good, just a couple of things to take a look at.

api/src/main/webui/src/api/hooks/useResourceList.ts
MikeEdgar: Great catch! I think if `path` is added to the `queryKey`, that will give the isolation needed. The path always will have the cluster ID in it for context. I.e. `/api/kafkas/${kafkaId}/topics`.

MikeEdgar:
@alexcreasy I addressed all of your comments in the last commit and re-based the PR. Please take another pass.

api/src/main/webui/src/api/hooks/useResourceList.ts
alexcreasy: yeah that sounds like a good solution to isolate the cache!

alexcreasy (approved):
LGTM thanks Mike!

Issue #2682, opened 2026-06-24
Create automation for running system tests locally


alexcreasy:
Presently, the system tests are run via an action on PRs, but running the full suite locally requires a lot of prerequisites (deploying a specific local k8s cluster) before you can even run them.

This process needs to be simplified and automated for ease of development.

jankalinic:
Hi, just to add note to this task, systemtests can be run locally if you are on linux and use for example minikube - similar to what github actions do. The issue comes for MacOS users, since minikube runs inside a VM (or a container runtime like docker that itself runs in a Linux VM), which creates a network isolation layer between your Mac and the Kubernetes cluster. So you could technically "run it locally" using tests in a  container - test containers, that does not seem to be the right answer to this, since you want to step-debug tests preferably using IDE. More info can be found on https://minikube.sigs.k8s.io/docs/drivers/

alexcreasy:
Un-assigning myself and moving to backlog for now so as not to squat on the ticket. It's a little more complicated than initially thought and I don't have the time right now.

tomncooper:
@jankalinic How does #2685 tie into this?

tomncooper:
@MikeEdgar would #2696 help with this?

jankalinic:
The https://github.com/streamshub/console/pull/2685 is meant as overall documentation of STs that could help new people understand where is what and how it works. Not sure whether it fully achieves this issues goal. In particular, I'm wondering if "local + automation" is the right framing here, since the CI process is already automated and as a prequisite for any kubernetes operator is a kubernetes cluster (i think Mike and I are working on Minikube/Kind deployment scripts for the convenience). For manual ST execution, what tester really need is to know the required configuration (versions, etc.) for their specific debugging or testing scenarios.

MikeEdgar:
> would [#2696](https://github.com/streamshub/console/pull/2696) help with this?

@tomncooper that is the idea - one script to be used anywhere so that an environment is initialized for the `mvn` system tests.

MikeEdgar:
As noted on the PR to add an initial common script [1], this solution should include some more robust environment validations. E.g. checks for container environment (podman/docker), check for OPM (install if missing), etc. 

There is also some room to improve things by splitting out even more common functionality like the image building that is included in the new script. This could be used both locally as well as all the locations in the CI that need to build images.

[1] https://github.com/streamshub/console/pull/2696#discussion_r3499568804

PR #2696, opened 2026-06-29
Add common script to setup minikube for Playwright and system tests

MikeEdgar:
Add common script to setup minikube for Playwright and system tests in support of #2682.

In addition to adding the script, this PR modifies the `playwright-tests` and `systemtests` workflows to use it. Note, that `systemtests` will continue to use the version on `main` prior to this PR being merged. Once merged, future PRs will execute the `systemtests` workflow from the PR instead of from `main`. The Pre-Systemtests workflow will continue to execute from `main`, which is intended for security.

:construction: :warning: This is known to only work on Linux at the moment.

1. Run minikube setup
   ```shell
   ./systemtests/scripts/setup-minikube.sh
   ```

2. Run test(s)
   ```shell
   CONSOLE_CLUSTER_DOMAIN="$(minikube ip).nip.io" mvn -pl systemtests verify -DskipSTs=false -Dit.test=MessagesST
   ```

All commands above are from the project root.


systemtests/scripts/setup-minikube.sh
```diff
@@ -0,0 +1,123 @@
...
+    echo "Error occurred at line $1"
+    if [ -n "${SOCAT_PID}" ] ; then
+        kill ${SOCAT_PID}
+    fi
+    kill 0 # kill the master shell and all subshells
+}
+
+trap 'handle_error $LINENO' ERR
+
+OLM_VERSION="v0.42.0"
```
tomncooper: 0.45.0 is the most recent release

.github/workflows/systemtests.yml
```diff
@@ -56,14 +56,13 @@ jobs:
...
 
-
   # Run single or matrix of ST jobs
   run-systemtests:
     needs: generate-matrix
     strategy:
       fail-fast: false
       matrix: ${{ fromJson(needs.generate-matrix.outputs.matrix) }}
-      max-parallel: 2
+      #max-parallel: 2
```
tomncooper: Is this supposed to be commented out?

systemtests/scripts/setup-minikube.sh
```diff
@@ -0,0 +1,123 @@
...
+
+# Setup OLM -----------------------------------------------------
+OLM_INSTALL_SCRIPT=$(mktemp)
+
+curl --proto "=https" -o ${OLM_INSTALL_SCRIPT} -sL \
+  https://github.com/operator-framework/operator-lifecycle-manager/releases/download/${OLM_VERSION}/install.sh
+
+echo "${OLM_SCRIPT_SHA256} ${OLM_INSTALL_SCRIPT}" | sha256sum --check --status || { echo "Hash verification failed"; exit 1; }
+chmod +x ${OLM_INSTALL_SCRIPT}
+${OLM_INSTALL_SCRIPT} ${OLM_VERSION} || true
```
tomncooper, line 68: AFAIK the `||` will swallow any errors from the install script. I assume this is used as it would continue if OLM is already installed?

It might be better to do a proper check to see if it is installed and only install if not. In fact, it might be good to have a function to check all pre-requisites and other tests like the podman/docker check and env-var setup and do that first?

systemtests/scripts/setup-minikube.sh
```diff
@@ -0,0 +1,123 @@
...
+set -euox pipefail
+
+SOCAT_PID=""
+
+function handle_error {
+    echo "Error occurred at line $1"
+    if [ -n "${SOCAT_PID}" ] ; then
+        kill ${SOCAT_PID}
+    fi
+    kill 0 # kill the master shell and all subshells
```
tomncooper: That is a bit of a sledgehammer. You can use `jobs -p` to get the PIDs of process started by this script and only kill those. 

```shell
kill $(jobs -p) 2>/dev/null || true
```

systemtests/scripts/setup-minikube.sh
```diff
@@ -0,0 +1,123 @@
...
+}
+
+trap 'handle_error $LINENO' ERR
```
tomncooper: You might want to consider a cleanup function using EXIT which fires on every exit. Then you wouldn't need to have the socat kill at the end and would have all the clean up logic in one place.

systemtests/scripts/setup-minikube.sh
```diff
@@ -0,0 +1,123 @@
...
+ operator/target/bundle/streamshub-console-operator
+
+./operator/bin/generate-catalog.sh ./operator/target/bundle/streamshub-console-operator true
+
+${CONTAINER_RUNTIME} build \
+ -t localhost:5000/streamshub/console-operator-catalog:${PROJECT_VERSION} \
+ -f operator/src/main/docker/catalog.Dockerfile \
+ operator/
+
+socat TCP-LISTEN:5000,reuseaddr,fork TCP:$(minikube ip ${MK_PROFILE_ARG}):5000 &
```
tomncooper, line 125: Out of interest, why are we using this `socat` + `skopeo` dance instead of `minikube image load`?

systemtests/scripts/setup-minikube.sh
```diff
@@ -0,0 +1,123 @@
...
+SOCAT_PID=${!}
+# Wait for socat to warm up
+sleep 5
```
tomncooper: Can we not do something a bit more robust? Running on dev laptop vs GHA runner might make this brittle. 

Claude suggested pinging the TCP port until you get a response:
```shell
 for i in $(seq 1 30); do
      nc -z localhost 5000 && break
      sleep 1
  done
```

But there may be something more appropriate to `socat`?

.github/workflows/systemtests.yml
```diff
@@ -88,28 +87,33 @@ jobs:
...
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+
+      - name: Fetch Operator Framework Utilities
+        run: ./operator/bin/fetch-operator-utilities.sh
+
+      - name: Setup Minikube Environment
+        run: |
```
tomncooper, line 108: We do this setup in both playwrite-tests and here. Would it be better to put it in one place centrally, maybe another script like `ubuntu-mk-setup.sh` or a common action?

tomncooper (commented):
I think this is a good move. It would be good to continue this trend with the running of the system/playwrite tests so they could be run locally or in CI with a single command.

I ran the script locally on my Fedora 44 box with Minikube 1.38.1 and it completed successfully. I had a couple of suggested improvements.

My main question is why you didn't use `minikube image load` ?

.github/workflows/systemtests.yml
MikeEdgar: > Is this supposed to be commented out?

No, I intended to revert this. Another commit on the way.

systemtests/scripts/setup-minikube.sh
MikeEdgar: > I assume this is used as it would continue if OLM is already installed?

Exactly. I looked at the OLM install script and there are at least a couple of checks that it performs to see if OLM is already installed. I didn't want to duplicate that so opted to just ignore the exit code. It's certainly an ugly hack.

> It might be better to do a proper check to see if it is installed and only install if not. In fact, it might be good to have a function to check all pre-requisites and other tests like the podman/docker check and env-var setup and do that first?

I definitely agree with doing this, but for the first pass where the main target is GH actions with a known environment, maybe we can defer to the next PR for #2682 where the environment is less certain.

systemtests/scripts/setup-minikube.sh
MikeEdgar: I've been through several iterations and tried `image load` at one point without success. I'll try again and report back since it could have been some other issue in the script.

systemtests/scripts/setup-minikube.sh
tomncooper: yeah, that is fair enough, we can defer the belt and braces approach as we still mainly in GHA

systemtests/scripts/setup-minikube.sh
MikeEdgar: I will check this also. The origin of the `socat` stuff is from the minikube documentation on this page: https://minikube.sigs.k8s.io/docs/handbook/registry/

.github/workflows/systemtests.yml
MikeEdgar: I agree and I also suspect we will want to split out the image building from the minikube creation so it's reusable from other actions (e.g. snapshot.yml where we build/push to Quay on each `main` commit).

systemtests/scripts/setup-minikube.sh
MikeEdgar: I'll update this. We updated to 0.42.0 relatively recently (May). These OLM guys are too fast!

systemtests/scripts/setup-minikube.sh
MikeEdgar: Let me check this. This handling is from other scripts we have and I thought the benefit of trapping `ERR` was to get the line number of an error. If we can have a single or combined trap, I agree that's better.

systemtests/scripts/setup-minikube.sh
MikeEdgar: Loading the images works for me locally. I believe it's putting them in the minikube host's local docker images versus the registry addon, so as long as none of the resources attempt to pull, it should work. The pull policies would need to be `IfNotPresent`. I'll test a bit more with it.

MikeEdgar: @tomncooper OLM creates the catalog pod with an `Always` pull policy and I don't see a way to override that. This might be a battle for another day.

systemtests/scripts/setup-minikube.sh
MikeEdgar: I think it could still leave zombies for sub-sub-processes (spawned by children), but probably good enough here. I'll include this in the `EXIT` trap.

systemtests/scripts/setup-minikube.sh
```diff
@@ -0,0 +1,138 @@
...
+    # Capture the exact exit code of the last executed command
+    local exit_code=${?}
+
+    echo "🧹 Cleaning up..."
+    if [ -e ${OLM_INSTALL_SCRIPT} ] ; then
+        rm -vf ${OLM_INSTALL_SCRIPT}
+    fi
+
+    if [ "${exit_code}" -ne 0 ]; then
+        echo "❌ Failed with exit code ${exit_code} at line ${last_line}"
```
tomncooper: Unless there is an error `last_line` is undefined. Which wouldn't be an issue unless you can somehow get a non-zero exit code _without_ an error, which I think SIGTERM does? Anyway, probably best to initalize `last_line` to something.

.github/workflows/systemtests.yml
```diff
@@ -122,16 +126,13 @@ jobs:
...
 
-          gh run download ${{ env.ARTIFACTS_RUN_ID }} --repo ${{ env.REPO }} -n k8s-resources --dir ./k8s-resources
-
-          mkdir -p ./k8s-resources/merged
-          cat ./k8s-resources/kubernetes/consoles.console.streamshub.github.com-v1.yml \
-              ./k8s-resources/kubernetes/kubernetes.yml \
-              > ./k8s-resources/merged/streamshub-console-operator.yaml
+          cat ./operator/target/kubernetes/consoles.console.streamshub.github.com-v1.yml \
+              ./operator/target/kubernetes/kubernetes.yml \
+              > ./target/streamshub-console-operator.yaml
```
tomncooper, line 131: Does the `./target/` definitely  exist? You have a `mkdir` in the old command, do you need one here too?

systemtests/scripts/setup-minikube.sh
```diff
@@ -0,0 +1,138 @@
...
+ -f operator/src/main/docker/catalog.Dockerfile \
+ operator/
+
+socat TCP-LISTEN:5000,reuseaddr,fork TCP:$(minikube ip ${MK_PROFILE_ARG}):5000 &
+
+# Wait for socat to warm up
+for i in $(seq 1 30); do
+    nc -z localhost 5000 && break
+    sleep 1
+done
```
tomncooper, line 131: Nit: If this fails 30 times in a row, the script just continues. You could throw a final check in to catch this:
```shell
nc -z localhost 5000 || { echo "socat not reachable"; exit 1; }
```

tomncooper (approved):
LGTM, just a couple of nits/suggestions.

My only thought was do you need to also delete the old `.github/actions/setup-minikube/action.yml`?

.github/workflows/systemtests.yml
MikeEdgar: Yes, the `mvn` command executed in the `setup-minikube.sh` script will have created it.

MikeEdgar:
@tomncooper your latest suggestions are applied in the final commit. I'll merge the PR once the CI passes.

PR #2719, opened 2026-07-04
Topic message table fixes and alignment with previous Next.js version

### LINKED ISSUE #2716: [UI] Messages page - Hidden filter dropdown
**Describe the bug**
When there are few or 0 message results in messages page, the dropdown filter element is hidden, but the results element now has a scroll bar on the right.

**To Reproduce**
1. Go to messages page
2. Set filter where 1 or 0 messages meets the query
3. Open the filter dropdown
4. See error that the dropdown does not fit the element and it's hidden

**Expected behavior**
Even when there are no results - the element should have full - possible height

**Screenshots**

Image

### LINKED ISSUE #2717: [UI] Messages page - timestamp query not working + hides the searchbar
**Describe the bug**
When any timestamp query is searched on the messages page, the search bar disappears and page displays No messages data. The only way to get out of the error, is to enter any other page and then navigate to messages page again. Tried with the dropdown timestamp just as with the query - both result in the error. 

Example: 
Message with offset 49 and timestamp 2026-07-03 21:53:32Z exists in the table before search.
When searched (using UI or directly via query) for timestamps `2026-07-03T22:00` or `2026-07-03T20:00` or `2026-07-03T23:00`. The error message appears.

**To Reproduce**
1. Go to messages page
2. Fill the search bar with a timestamp query
3. Click search

**Expected behavior**
The result message to be shown, when correct timestamp is selected.
Even with incorrect query, the messages page should display the search bar. 
The results element should be the only place where the error is displayed without hiding the search bar.

**Screenshots**

Image

url contains ``

Image

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser chrome
 - Version 0.13.0-snapshot

MikeEdgar:
Fixes #2716
Fixes #2717

- Do not hide search form for fetch errors or no messages
- Fix handling of timestamp search criterion
- Fix message drawer to overlay table and not scroll when table scrolls. This allows messages further down in the table to be selected and viewed.
- Fix position and size of the advanced message search box
- Display message headers as key/value pair rather than as a JSON object
- Add back advanced JSON display for key and value shown in the message drawer. Add support for dark mode which was not present in the old Next.js version

Assisted-by: IBM Bob

### DIFF
--- a/api/src/main/webui/src/components/kafka/topics/AdvancedSearch.tsx
+++ b/api/src/main/webui/src/components/kafka/topics/AdvancedSearch.tsx
@@ -3,7 +3,7 @@
  * Provides filtering options for Kafka messages
  */
 
-import { useState, useCallback, useTransition, useRef, useEffect } from 'react';
+import { useState, useCallback, useTransition, useRef, useLayoutEffect } from 'react';
 import { createPortal } from 'react-dom';
 import { useTranslation } from 'react-i18next';
 import { toDatetimeLocal, toISOWithLocalOffset } from '@/utils/dateTime';
@@ -195,7 +195,7 @@ export function AdvancedSearch({
     }
   }, []);
 
-  useEffect(() => {
+  useLayoutEffect(() => {
     if (!isExpanded) return;
     updatePanelPosition();
     window.addEventListener('scroll', updatePanelPosition, true);
--- a/api/src/main/webui/src/components/kafka/topics/AdvancedSearch.tsx
+++ b/api/src/main/webui/src/components/kafka/topics/AdvancedSearch.tsx
@@ -3,8 +3,10 @@
  * Provides filtering options for Kafka messages
  */
 
-import { useState, useCallback, useTransition } from 'react';
+import { useState, useCallback, useTransition, useRef, useEffect } from 'react';
+import { createPortal } from 'react-dom';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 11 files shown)


jankalinic (approved):
Thanks, nothing stands out to me, approved.

api/src/main/webui/src/components/kafka/topics/AdvancedSearch.tsx
```diff
@@ -172,8 +176,38 @@ export function AdvancedSearch({
...
+        left: rect.left,
+        width: rect.width,
+        maxHeight: `calc(100vh - ${top}px - 16px)`,
+        overflowY: 'auto',
+        zIndex: 9999,
+      });
+    }
+  }, []);
+
+  useEffect(() => {
```
alexcreasy: This would be better as a useLayoutEffect it stops some random flashes when doing direct dom manipulation.

alexcreasy (changes requested):
One small change requested! LGTM otherwise!

api/src/main/webui/src/components/kafka/topics/AdvancedSearch.tsx
MikeEdgar: Updated `useEffect` with `useLayoutEffect`.

alexcreasy (approved):
LGTM, thanks for the change!

PR #2737, opened 2026-07-08
Escape commas in `like` searches from UI

### LINKED ISSUE #2733: [UI] Group filter not working with special chars
**Describe the bug**
The Groups page name-filter works normally with most of the characters, after a while of testing i found that
there are characters like `?` `*` `,` that do not return the correct groups - either return all groups or none.

**To Reproduce**
1. Go to Groups page
2. Fill in the name filter with any of the mentioned characters
3. See error

**Expected behavior**
Filter working with all allowed consumer group name characters.

**Screenshots**
Original
Image

Not working chars:
Image

Image

Image

and for example working char

Image

### LINKED ISSUE #2759: [ST] Enable user group test case (comma separated)
Currently disabled test for // Map.entry("Comma separated", "group,comma,separated,,"), once https://github.com/streamshub/console/pull/2737 is merged, enable in tests

MikeEdgar:
Allow commas to be used in search forms. This requires escaping the `,` character in the parameter sent from the UI to the API so that it is not used as a delimiter in the API's query parsing.

Fixes #2733

Assisted-by: IBM Bob

### DIFF
--- a/api/src/main/java/com/github/streamshub/console/api/model/FetchFilter.java
+++ b/api/src/main/java/com/github/streamshub/console/api/model/FetchFilter.java
@@ -1,6 +1,9 @@
 package com.github.streamshub.console.api.model;
 
+import java.util.ArrayList;
 import java.util.List;
+
+import com.github.streamshub.console.api.support.FetchFilterPredicate;
 
 public class FetchFilter {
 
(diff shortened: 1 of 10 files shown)


jankalinic (approved):

MikeEdgar:
I will hold off merging this one until 0.13.0 is released. I would also like to update the relevant system test.

MikeEdgar:
/systemtests run --profile=regression --install-type=olm --retry-count=0 --env=TEST_CONSOLE_LOG_LEVEL=DEBUG --testcase=GroupsST

MikeEdgar:
The system tests in `GroupsST` actually all passed, but this error occurred in post-processing:

> 2026-07-14T21:01:12.5575642Z ##[error]The path for one of the files in artifact is not valid: /screenshots/2026-07-14__20-07-44/Type: DateTime ISO - DateTime: ISO_8601_DateTime - Offset: "0"/click-success-2026-07-14__20-40-23-859.png. Contains the following character:  Double quote "
>          
> Invalid characters include:  Double quote ", Colon :, Less than <, Greater than >, Vertical bar |, Asterisk *, Question mark ?, Carriage return \r, Line feed \n
>          
> The following characters are not allowed in files that are uploaded due to limitations with certain file systems such as NTFS. To maintain file system agnostic behavior, these characters are intentionally not allowed to prevent potential problems with downloads on different file systems.

jankalinic:
Good catch @MikeEdgar . I can create an issue for that and fix it. But keeping this approved, so you can merge at anytime.

PR #2763, opened 2026-07-14
[ST] Tests can handle creating catalogsource

jankalinic:
In past, user had to deploy their own catalogsource for running OLM tests, now they can pass a value if they want to let STs handle this operation. If no image is specified or catalog exists, tests simply skip creating the catalogsource.


MikeEdgar (approved):
Nice!

PR #2776, opened 2026-07-16
[ST] Enable AuthST, fix several tests, performance improvements

jankalinic:


systemtests/src/main/java/com/github/streamshub/systemtests/utils/playwright/PwPageUrls.java
```diff
@@ -43,7 +43,8 @@ public static String getMessagesPage(TestCaseConfig tcc, String kafkaName, Strin
     }
 
     public static String getGroupsMembersPage(TestCaseConfig tcc, String kafkaName, String consumerGroupName) {
-        return getGroupsPage(tcc, kafkaName) + "/" + consumerGroupName + "/members";
+        consumerGroupName = consumerGroupName.isBlank() ? "" : "/" + consumerGroupName;
```
MikeEdgar: I'm curious about this being blank. If that happens, it looks like this would end up with a URL like `kafka/${kafkaId}/groups/members`. Are you seeing that happening?

jankalinic: I navigate to it with groupName="" at the end of the test, and try to check if those are visible for the user with no access, if its blank the url ends up {kafka}/groups//members which is an incorrect url that shows just Error. With {kafka}/groups/members i assert that this page is forbidden for the user without the correct rights. Or does it work differently, should i check that any other way?

MikeEdgar: > if its blank the url ends up {kafka}/groups//members which is an incorrect url that shows just Error.

That seems fine.

> With {kafka}/groups/members i assert that this page is forbidden for the user without the correct rights.

For that URL, I would guess it's trying to find a group with an ID encoded as "members", which could either give a 403 or a 404 depending on the security configuration.

jankalinic: I'm trying to test the negative path - user with only view topics cannot view groups. So in that case i'm expecting 403. Is that correct way to do it? Or would you suggest trying to reach out to specific groupId and then asserting 403 response?

MikeEdgar: If they only have view topics, they could get a 403 listing the groups, and also if you want to try and view a single group that you know exists, that will give 403 as well.

jankalinic: So do you consider this verification ok, or would you like me to change it?

MikeEdgar: It's technically ok, but it might obscure the intent by using a blank group ID as a way to fetch a specific group named "members".

jankalinic: Changed to target specific groupId members and removed the blank conditional text for url.

jankalinic:
/systemtests run --testcase=AuthST --install-type=olm --retry-count=2

jankalinic:
/systemtests run --testcase=AuthST --install-type=olm --retry-count=2

jankalinic:
/systemtests run --testcase=AuthST --install-type=olm --retry-count=2

MikeEdgar:
@jankalinic I'm looking into this. I may expand the scope to make sure all of the tests are running OK.

jankalinic:
Thank you @MikeEdgar. Feel free to modify or decide the future of this PR.

systemtests/scripts/setup-minikube.sh
```diff
@@ -44,6 +44,7 @@ if ! minikube status ${MK_PROFILE_ARG} >/dev/null 2>&1 ; then
     --driver=${MINIKUBE_DRIVER:-kvm2} \
     --cpus=${MINIKUBE_CPU_COUNT:-6} \
     --memory=${MINIKUBE_MEMORY:-16384} \
+    --disk-size=${MINIKUBE_DISK_SIZE:-50g} \
```
MikeEdgar, line 47: The default disk allocation in minikube is insufficient for the Kafka Connect image build process.

systemtests/src/main/java/com/github/streamshub/systemtests/interfaces/TestBucketExtension.java
```diff
@@ -91,7 +91,11 @@ public void beforeTestExecution(ExtensionContext testMethodContext) throws Excep
                     method.getAnnotation(SetupTestBucket.class).value().equals(testBucketGroupName)) {
                     // Called method needs to be either explicitly public or needs to be -> `method.setAccessible(true)`
                     LOGGER.info("Invoking @SetupTestBucket method [{}] for TestBucket [{}]", method.getName(), testBucketGroupName);
-                    method.invoke(testMethodContext.getRequiredTestInstance());
+                    if (method.trySetAccessible()) {
+                        method.invoke(testMethodContext.getRequiredTestInstance());
+                    } else {
+                        throw new IllegalStateException("Unable to access @SetupTestBucket method " + method);
+                    }
```
MikeEdgar, line 98: This allows us to keep the test classes as package-private (JUnit best practice) with minimal fuss.

systemtests/src/main/java/com/github/streamshub/systemtests/logs/LogWrapper.java
```diff
@@ -30,18 +30,18 @@ private static void forceLoggingConfiguration() {
         ConsoleAppender consoleAppender = ConsoleAppender.newBuilder()
                 .setName(STDOUT)
                 .setLayout(PatternLayout.newBuilder()
-                    .withPattern("%d{yyyy-MM-dd HH:mm:ss}{GMT} [%thread] %highlight{%-5p} [%c{1}:%L] %m%n")
+                    .setPattern("%d{yyyy-MM-dd HH:mm:ss}{GMT} [%thread] %highlight{%-5p} [%c{1}:%L] %m%n")
```
MikeEdgar, line 33: These `with` methods are deprecated in favor of the `set` equivalents.

systemtests/src/main/java/com/github/streamshub/systemtests/setup/console/OlmConfig.java
```diff
@@ -120,7 +120,7 @@ public void delete() {
     public OlmVersionModificationData getUpgradeData() {
         maybeCreateCatalogSource();
 
-        String newVersion = Environment.getConsoleOperatorVersion();
+        String newVersion = Environment.getConsoleOperatorVersion(String::toLowerCase);
```
MikeEdgar, line 123: The version from the POM needs to be lower-cased for the OLM upgrade where the images and OLM artifacts use lowercase `-snapshot` as the suffix.

systemtests/src/main/java/com/github/streamshub/systemtests/utils/resourceutils/kafka/KafkaTopicUtils.java
```diff
@@ -365,32 +395,16 @@ public static void setupUnmanagedUnderReplicatedAndUnavailableTopics(String name
         KafkaUtils.removeAnnotation(namespace, kafkaName, ResourceAnnotations.ANNO_STRIMZI_IO_SKIP_BROKER_SCALEDOWN_CHECK, true);
     }
 
-    private static KafkaClients buildClients(String namespace, String kafkaName, String kafkaUser, String topicName, int messageCount) {
-        return new KafkaClientsBuilder()
```
MikeEdgar, line 369: I think we can (over time) migrate the `KafkaClients` and Strimzi kafka client image Jobs to just use the Kafka Admin, Consumer, and Producer clients instead. It's significantly faster and minimizes the number of moving parts in the tests.

systemtests/src/main/java/com/github/streamshub/systemtests/utils/Utils.java
```diff
@@ -148,4 +150,25 @@ public static void sleepWait(long timeInMilis) {
             LOGGER.error("Sleep was interrupted due to: {}", e.getMessage());
         }
     }
+
+    public static CompletableFuture<Void> runAsyncWithContext(Runnable runnable) {
```
MikeEdgar, line 154: These methods allow us to run more of the test setup concurrently on worker threads while using the same test context. The Kubetest4j libraries relies on `ThreadLocal` a lot, so copying the context across threads is necessary.

systemtests/src/test/java/com/github/streamshub/systemtests/auth/AuthST.java
```diff
@@ -393,31 +421,71 @@ void testClassSetup() {
...
+
+    @BeforeEach
+    void testCaseSetup() {
+        for (var clusterName : List.of(AuthTestConstants.TEAM_DEV_KAFKA_NAME, AuthTestConstants.TEAM_ADMIN_KAFKA_NAME)) {
+            try (Admin admin = KafkaClientsUtils.createSecureClient(
+                    tcc.namespace(),
+                    clusterName, 
+                    KafkaNamingUtils.kafkaUserName(clusterName),
+                    Admin::create)) {
+                Wait.until("Groups deleted", 2_000, 120_000, () -> {
```
MikeEdgar, line 470: This long timeout is needed because the test clients Job appears to not call `close` on consumers. Only after a timeout does Kafka transition the groups from `Stable` to `Empty`, required for deletion. This won't be needed if/when we migrated to just using Producers/Consumers in-process.

systemtests/src/test/java/com/github/streamshub/systemtests/kafka/ConnectST.java
```diff
@@ -197,60 +204,94 @@ void testFilterKafkaConnectClustersAndConnectors() {
...
+                    tcc.kafkaUserName(),
+                    tcc.consoleInstanceName());
+        };
+
+        if (Environment.CONNECT_DEPLOY_CONCURRENT) {
+            LOGGER.info("CONNECT_DEPLOY_CONCURRENT is true, deploying {} and {} concurrently", KAFKA_CONNECT_SRC_NAME, KAFKA_CONNECT_SINK_NAME);
+            connect = CompletableFuture.allOf(srcConn, runAsyncWithContext(deployConnectSink));
+        } else {
+            LOGGER.info("CONNECT_DEPLOY_CONCURRENT is false, deploying {} and {} sequentially", KAFKA_CONNECT_SRC_NAME, KAFKA_CONNECT_SINK_NAME);
+            connect = srcConn.thenComposeAsync(runComposableAsyncWithContext(deployConnectSink));
```
MikeEdgar, line 235: Apparently, GitHub runners do not have enough disk space for the concurrent connect image builds.

systemtests/src/test/java/com/github/streamshub/systemtests/upgrade/OlmUpgradeST.java
```diff
@@ -68,7 +68,7 @@ void testUpgradeOlmOperator() {
         ConsoleOperatorSetup consoleOperatorSetup = new ConsoleOperatorSetup(Constants.CO_NAMESPACE);
         consoleOperatorSetup.setInstallConfig(olmConfig);
         LOGGER.info("Installing Console Operator via OLM into namespace '{}'", Constants.CO_NAMESPACE);
-        consoleOperatorSetup.install(false);
+        consoleOperatorSetup.install(true);
```
MikeEdgar, line 71: When this test runs with other test classes, any previously-installed operator needs to be removed first, false -> true

systemtests/src/main/java/com/github/streamshub/systemtests/utils/resourceutils/ClusterUtils.java
```diff
@@ -29,27 +36,39 @@ public static void checkClusterHealth() {
...
-        LOGGER.trace("Cluster type detected as {}", isOcp ? "OpenShift" : "Kubernetes");
-        return isOcp;
+        return (Boolean) CACHE.computeIfAbsent("isOCP", k -> {
+            boolean isOcp = KubeResourceManager.get().kubeCmdClient().exec(false, false, "api-versions").out().contains("openshift.io");
+            LOGGER.info("Cluster type detected as {}", isOcp ? "OpenShift" : "Kubernetes");
+            return isOcp;
+        });
     }
 
     public static String getClusterDomain() {
```
MikeEdgar, line 46: This reworks cluster domain lookup now respects `CONSOLE_CLUSTER_DOMAIN` primarily, then if not set it will derive from the OCP base domain as before (if OpenShift), or try to use the Kubernetes master URL's IP address to construct a `.nip.io` domain. This should help in local environments where, for example, minikube's IP is changing each time the cluster is created.

systemtests/scripts/setup-minikube.sh
MikeEdgar: This change has no effect in GitHub actions, but is useful for local testing.

MikeEdgar:
@jankalinic please take a look next week and let me know if I've done anything objectionable with this. Aside from make some of the tests pass, this PR now has some performance improvements and a start to migrating away from the kafka clients jobs to just using the Kafka clients directly.

jankalinic (commented):
WOW. I'm speechless. There's a lot of new approaches to things. I reviewed it and as i see tests are passing, cannot make any point why this should not be merged. I'm surprised especially by the kafka clients and also the concurrent setup.

MikeEdgar (approved):

PR #2783, opened 2026-07-21
Configure operator lease using builder

### LINKED ISSUE #2647: Migrate from LeaderElectionConfiguration subclass to properties
We currently have a subclass of `LeaderElectionConfiguration` [1] to configure the name of the operator lease. The constructors on the super class have been deprecated for removal, so we should (attempt to) migrate to using configuration properties [2] instead. I believe we can accomplish the same by setting these in the `application.properties` of the operator.

```
josdk.leader-election.enabled=true
josdk.leader-election.lease-name=streamshub-console-operator-lease
```

We do not currently have a test checking the lease, so a simple one can be added checking for the creation of a `coordination.k8s.io/leases` resource with the expected name.

[1] https://github.com/streamshub/console/blob/ff8309baf25cfb0921b2d7ea1cbbf97d5d4a6f34/operator/src/main/java/com/github/streamshub/console/LeaderConfiguration.java
[2] https://javaoperatorsdk.io/docs/documentation/operations/configuration/#operator-level-configuration-keys

MikeEdgar:
Closes #2647 

---

As noted in the issue, the JOSDK properties have no effect. Instead, this PR will use a CDI producer method that provides the `LeaderElectionConfiguration` using the associated builder. This avoids using the deprecated constructors that were used by sub-classing it previously.


operator/src/test/java/com/github/streamshub/console/ConsoleReconcilerTest.java
```diff
@@ -1460,4 +1461,13 @@ void testConsoleReconciliationWithResourceConflicts() {
             assertThat(errorMessages, hasItem(containsString(CONSOLE_NAME + '-' + PrometheusDeployment.NAME)));
         });
     }
+
+    @Test
+    void testOperatorLeaseCreation() {
+        var lease = client.resources(Lease.class).withName("streamshub-console-operator-lease").get();
```
tomncooper:
```suggestion
        var lease = client.resources(Lease.class).withName(LeaderConfigurationFactory.LEASE_NAME).get();
```

tomncooper (approved):
LGTM. One Nit.

operator/src/test/java/com/github/streamshub/console/ConsoleReconcilerTest.java
MikeEdgar: Ah, yes. Good catch, this was 50% of the motivation for creating the constant anyway :rofl:

PR #2795, opened 2026-07-23
Prevent mixed content redirect following successful Kafka cluster login

### LINKED ISSUE #2728: [UI] Modal does not go away after entering correct kafka credentials
**Describe the bug**
When kafka credentials auth method is used for console login, there's a modal that prevents user from seeing the page saying "authentication failed".Once user logs in with correct credentials. When this page is manually reloaded, the modal goes away.

**To Reproduce**
1. Go to console login (kafka credentials setup)
2. Log in with correct credentials
3. See error

**Expected behavior**
Login should succeed with correct credentials and let user without reloading see the page.

**Screenshots**

Image

**Desktop (please complete the following information):**
 - OS: N/A
 - Browser [chrome
 - Version 0.13.0-snapshot

MikeEdgar:
Fixes #2728 

The problem seen in the tests was due to a redirect being issued by Quarkus to a default `/login.html`. The protocol in the redirect URL is `http`, but the tests use the Console via an ingress with TLS/HTTPS enabled. In this case, the browser rejects the login response because it is "mixed content" - an HTTPS response that contains an HTTP redirect header. The solution here is to disable the redirect entirely (we don't need it anyway) when building the form authentication mechanism for the Kafka cluster.

This PR also includes a small change in the system tests to ensure that the Kube cluster domain is provide for non-OCP scenarios.

### DIFF
--- a/systemtests/src/main/java/com/github/streamshub/systemtests/utils/playwright/PwUtils.java
+++ b/systemtests/src/main/java/com/github/streamshub/systemtests/utils/playwright/PwUtils.java
@@ -17,10 +17,13 @@ import com.microsoft.playwright.Browser;
 import com.microsoft.playwright.BrowserContext;
 import com.microsoft.playwright.BrowserType;
 import com.microsoft.playwright.Locator;
+import com.microsoft.playwright.Locator.GetByRoleOptions;
 import com.microsoft.playwright.Page;
 import com.microsoft.playwright.Playwright;
 import com.microsoft.playwright.TimeoutError;
 import com.microsoft.playwright.Tracing;
+import com.microsoft.playwright.assertions.PlaywrightAssertions;
+import com.microsoft.playwright.options.AriaRole;
 import com.microsoft.playwright.options.WaitForSelectorState;
 import com.microsoft.playwright.options.WaitUntilState;
 import io.fabric8.kubernetes.api.model.Secret;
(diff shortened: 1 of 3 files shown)


MikeEdgar:
/systemtests run --testcase=KafkaCredentialsST --install-type=olm --retry-count=2

systemtests/src/main/java/com/github/streamshub/systemtests/utils/playwright/PwUtils.java
```diff
@@ -702,12 +705,17 @@ public static void loginWithKafkaCredentials(TestCaseConfig tcc, String namespac
...
-        waitForLocatorAndFill(tcc, CssSelectors.PAGES_KAFKA_CREDENTIALS_PASSWORD_INPUT, password);
-        waitForLocatorAndClick(tcc, CssSelectors.PAGES_KAFKA_CREDENTIALS_LOGIN_BUTTON);
-        Utils.sleepWait(TimeConstants.UI_COMPONENT_REACTION_INTERVAL_SHORT);
-        // TODO: remove once fixed
-        PwUtils.reload(tcc);
+        var loginModal = tcc.page().getByRole(AriaRole.DIALOG);
+        PlaywrightAssertions.assertThat(loginModal).isVisible();
+        loginModal.locator("form input#username").fill(kafkaUser);
+        loginModal.locator("form input#password").fill(password);
+        loginModal.getByRole(AriaRole.BUTTON, new GetByRoleOptions().setName("Login")).click();
```
MikeEdgar, line 712: These seem to be so much quicker when I use these locally. What was the issue you saw with them @jankalinic ?

jankalinic: There were a few reasons that i had in mind while creating the css selectors. 

Repeatability - using constants for easier fixes - well, that did not work out :D. 
Duplicates - once .locator() spots anything that's not unique (e.g. 3 buttons, labels, text inputs - without a proper element#id) it crashes and good luck finding out which selector was the failing one and what are the others. 
Unification - by locating them all the same way it seemed more natural, uncomplicated, in place. 
First testing also showed some flakiness - using the AriaRole e.g. Now i cannot remember what exactly was the issue, so i cannot say it with confidence.

tldr;
I am not against using those locators shown in the diff, but i'd like to solve it in a systematic way, where we won't have to update the css parts like at all.... or atleast not like it's done now - you know how painful that is :D . I know for some things the OUIA, testID and other helpers would be best, for the rest we could simplify to your proposed solution while also avoiding the "3-body" problem of element uniqueness in all of the situations with different action states (dropdowns, clicked, hovered etc). 

Also the .fill(), i think we removed that, tested on minikube and it worked perfectly, then i tested that on ocp (showing as flaky) and had to put it back with the retry mechanism. Again, i'd love to use the ones in this diff as it's simple, does not add any complexity to the code and "just works".

jankalinic (approved):

PR #2796, opened 2026-07-24
[ST] Yaml upgrade test - use local built bundle as HEAD

### LINKED ISSUE #2766: [ST] Operator upgrade tests - use HEAD as lastest operator version
Currently we operator update test resource files with urls and set versions that need to be manually changed each release. Might be better for OLM to get the version dynamically. YamlUpgrade.yml urls, pointing to latest released, need to use version in development main->HEAD and avoid waiting for the release.

jankalinic:
YamlUpgradeST used to get its old/new Console Operator version and CRDs bundle URL from a dedicated YamlUpgrade.yaml file. This PR drops the YAML file and puts everything into Environment.java in ENV format.

- OLD_CONSOLE_OPERATOR_VERSION / OLD_CONSOLE_OPERATOR_CRDS_URL — the version to upgrade from
- NEW_CONSOLE_OPERATOR_CRDS_URL — optional override for the version to upgrade to. When unset its derived from local operator file from running `mvn` build.
- removed oldtopics ui selectors and test checks
- a change in the github workflow gate was made to ensure Environment old operator version stays updated.

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -16,19 +16,20 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Ensure upgrade YAMLs changed
+      - name: Ensure old operator version was bumped
         run: |
-          CHANGED=$(git diff --name-only \
+          CHANGED=$(git diff \
             ${{ github.event.pull_request.base.sha }} \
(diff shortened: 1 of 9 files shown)


MikeEdgar (approved):
The process looks good.

PR #2805, opened 2026-07-27
[WIP][ST] Remove fixed css selectors, utilize components from PW and OUIA ids

jankalinic:
This PR does following:

* adds a new ouia ids in the api/webui for a certain elements that are checked during the tests
* refactors CssSelectors so instead of using only constants with full css path it uses Roles, element types, ids, ouia etc.

The changes made in this PR get rid of the long CssSelectors and it's classes, but keeps the overall logic of -> for groups page in tests - get selectors from GroupsPage..... for topics page from TopicsPage etc, so that it's clear where lies what. Adding some common constants used throughout the tests and selectors, separated OUIA constants for easier lookup.

This refactor needs followup PRs with changes touching the PwUtils, but it should work without it as well.

Assisted by: Sonnet 5


jankalinic:
/systemtests run --profile=regression --install-type=olm --retry-count=2

jankalinic:
/systemtests run --profile=regression --install-type=olm --retry-count=2

jankalinic:
/systemtests run --profile=regression --install-type=olm --retry-count=2

PR #2806, opened 2026-07-27
[ST] Improve name sanitization

jankalinic:
Allow only letters, digits and underscore + limit the max length of the screenshot name


MikeEdgar (approved):
👍

PR #2918, opened 2026-09-10
Restrict CI checks to changes that effect code only

tomncooper:
Currently, the CI will run all tests for _any_ change in the repo, even for changes to files that have no effect on the operation of the Console or its operator (like docs updates). 

This PR adds ignore paths for non-functional files and folders in the repo so that changes to them won't trigger the testing pipelines.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -4,12 +4,26 @@ on:
     branches:
       - 'main'
       - '[0-9]+.[0-9]+.x'
+    paths-ignore:
+      - 'docs/**'
+      - '*.md'
+      - 'LICENSE'
+      - 'dco.txt'
+      - '.gitignore'
+      - '.snyk'
   pull_request:
     branches:
       - 'main'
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yml
```diff
@@ -4,12 +4,28 @@ on:
     branches:
       - 'main'
       - '[0-9]+.[0-9]+.x'
+    paths-ignore:
+      - 'docs/**'
+      - 'examples/**'
```
MikeEdgar: There are some resources in `examples` that are used by the Playwright tests. We should probably exclude them from this list only because of that.

MikeEdgar (changes requested):
Just one comment about the Playwright tests.

MikeEdgar (approved):

MikeEdgar:
Unfortunately I don't think we can have conditionally required workflows in the GH configuration. So when we have PRs like this, we'll have to bypass the rules and merge them.

